PR #2, opened 2019-12-11
Use tox to be able to run tests locally

toabctl:
With tox, it's easy to run unittests or linters locally before
submitting code. So use it here.
Also switch Travis to use tox, too. That way the results between local
test runs and travis runs are more compareable.


rjfd (approved):
lgtm

rjfd:
@toabctl thanks for this contribution.

PR #27, opened 2020-01-15
core: Check minion FQDN not resolving to loopback IP address

### LINKED ISSUE #21: MGRs are using the localhost network as in  `127.0.0.1:6800`
```
$ ceph mgr dump
...
    "active_addrs": {
        "addrvec": [
            {
                "type": "v2",
                "addr": "127.0.0.1:6800",
                "nonce": 1
            },
            {
                "type": "v1",
                "addr": "127.0.0.1:6801",
                "nonce": 1
            }
        ]
    },
```
We should make sure, MGRs use proper addresses.

ricardoasmarques:
Fixes: https://github.com/SUSE/ceph-bootstrap/issues/21

### DIFF
--- a/ceph_bootstrap/exceptions.py
+++ b/ceph_bootstrap/exceptions.py
@@ -6,3 +6,9 @@ class CephNodeHasRolesException(CephBootstrapException):
     def __init__(self, minion_id, roles):
         super(CephNodeHasRolesException, self).__init__(
             "Cannot remove host '{}' because it has roles defined: {}".format(minion_id, roles))
+
+
+class CephNodeFqdnResolvesToLoopbackException(CephBootstrapException):
+    def __init__(self, minion_id):
+        super(CephNodeFqdnResolvesToLoopbackException, self).__init__(
+            "Host '{}' FQDN resolves to the loopback interface IP address".format(minion_id))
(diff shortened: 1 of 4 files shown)


ceph_bootstrap/exceptions.py
```diff
@@ -6,3 +6,9 @@ class CephNodeHasRolesException(CephBootstrapException):
     def __init__(self, minion_id, roles):
         super(CephNodeHasRolesException, self).__init__(
             "Cannot remove host '{}' because it has roles defined: {}".format(minion_id, roles))
+
+
+class CephNodeHasNoPublicIPv4Exception(CephBootstrapException):
+    def __init__(self, minion_id):
+        super(CephNodeHasNoPublicIPv4Exception, self).__init__(
+            "No public IPv4 address configured on host '{}'".format(minion_id))
```
rjfd: I think the error message should be:
`"Host '{}' FQDN resolves to the loopback interface IP address"`

and the exception class name should be `CephNodeFqdnResolvesToLoopbackException`

rjfd (changes requested):

ceph_bootstrap/exceptions.py
ricardoasmarques: Fixed

rjfd (approved):
lgtm

PR #33, opened 2020-01-18
qa: initial integration testing script

smithfarm:
The idea here is to have the ceph-bootstrap integration test code live in the same repo as ceph-bootstrap itself, and be installable on the target system (`zypper in ceph-bootstrap-qa`), and be easily runnable by both sesdev and teuthology.

Obviously, this is just a beginning. There's lots that could be added and improved, but I believe this meets the need of the hour.

NOTE: I suggest to review/try/test this PR together with https://github.com/SUSE/sesdev/pull/46

### DIFF
--- a/qa/README.rst
+++ b/qa/README.rst
@@ -1,0 +1,46 @@
+health-ok.sh
+============
+
+ceph-bootstrap integration test automation script
+
+
+Overview
+--------
+
+This bash script contains integration tests for validating Ceph deployments
+done with ceph-bootstrap.
+
+The idea is to run this script with the appropriate arguments on the
+Salt Master node after using ceph-bootstrap to deploy a cluster.
+
+The script makes a number of assumptions, as listed under "Assumptions", below.
+
+On success (HEALTH_OK is reached, sanity tests pass), the script returns 0.
+On failure, for whatever reason, the script returns non-zero.
+
+The script produces verbose output on stdout, which can be captured for later
+forensic analysis.
+
+Though referred to as a bash script, ``health-ok.sh`` is only the entry point.
+That file uses the bash internal ``source`` to run several helper scripts, which
+are located in the ``common/`` subdirectory.
+
+
+Assumptions
+-----------
+
+The script makes the following assumptions:
+
+1. the script is being run on the Salt Master of a Ceph cluster deployed using
+   ceph-bootstrap
+2. the script is being run as root
+3. the Ceph admin keyring is installed in the usual way, so the root user can
+   see and use the keyring
+
+
+Caveats
+-------
+
+The following caveats apply:
+
+1. Ceph will not work properly unless the nodes have (at least) short hostnames. That means the health-ok.sh script won't pass, either. There are two options: ``/etc/hosts`` or DNS
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -1,0 +1,205 @@
+#
+# This file is part of the ceph-bootstrap integration test suite
+#
+
+set -e
+
+# BASEDIR is set by the calling script
+source $BASEDIR/common/helper.sh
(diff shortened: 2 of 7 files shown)


qa/common/common.sh
```diff
@@ -0,0 +1,199 @@
+#
```
toabctl, line 1: Can't we just use python and a proper test framework (eg. being able to call just a single test, list available tests, mark slow running tests,...)?

smithfarm: We can if someone else does the work :-)

smithfarm: We could also turn the question around: instead of using Python and a complicated test framework, can't we just use a bash script? This PR is my answer to that question.

smithfarm: Please note that upstream has a Python-based testing framework called `teuthology` and the upstream developers have already written:

* a `cephadm` task for deploying Ceph with cephadm
* a `rados/cephadm` suite for testing core Ceph components in environments deployed using cephadm

So I think the need for integration testing via a complicated Python-based framework is already being met upstream, and we will be running those tests downstream in environments prepared by ceph-bootstrap.

qa/README.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Assumptions
+-----------
+
+The script makes the following assumptions:
+
+1. the script is being run on the Salt Master of a Ceph cluster deployed using
+   ceph-bootstrap
+2. the script is being run as root
+3. the Ceph admin keyring is installed in the usual way, so the root user can
```
ricardoasmarques, line 37: Does this mean that I have to manually install and configure `ceph client` on `Salt Master` to run QA tests?

smithfarm: Not sure what you mean by `ceph client` . . . I have tested this using

https://github.com/SUSE/sesdev/pull/46

and it "just works", no manual steps required.

ricardoasmarques: I have a `sesdev` env that I've created using the `--ceph-bootstrap-repo` option, and I got the following error when executing qa tests from `Salt Master`:

```
admin:~ # /root/ceph-bootstrap/qa/health-ok.sh
Running 'getopt --test'... PASS

WWWW: ceph_version_test
+ rpm -q ceph-common
package ceph-common is not installed
+ catch 1
+ echo

+ echo -n 'Overall result: '
Overall result: + '[' 1 = 0 ']'
+ echo 'NOT_OK (error )'
NOT_OK (error )
```

That's why I've asked if I have to manually install `ceph-common` on `Salt Master.`

smithfarm: @ricardoasmarques You have to use ceph-bootstrap to deploy a Ceph (SES7) cluster, first, before running `health-ok.sh`.

Maybe something has changed, but the last time I tried it, deploying SES7 caused `ceph-common` to be installed on the Salt Master. If this is no longer the case, this PR will need a modification.

rjfd: @smithfarm `ceph-common` is only installed in the minions with the `mon` role. Maybe you've tested in a single-node cluster where the salt-master is also a `mon`.

smithfarm: Does `ceph-bootstrap` have an "admin" role? My expectation is I can do `ceph -s` from the admin node.

ricardoasmarques: To run QA tests we need both `ceph-bootstrap` and `ceph` commands.
Given that `ceph-bootstrap` is installed on `Salt Master` and `ceph-common` is installed on `mon`s, maybe we have to execute ceph commands via ssh, e.g.: `ssh node1 ceph health`.

rjfd: No. The assumption is that you will have an admin keyring and ceph-common installed in all `mon` minions.

ricardoasmarques (approved):
lgtm

PR #50, opened 2020-01-29
salt_utils: redirect stdout to /dev/null when calling salt grains commands

### LINKED ISSUE #10: Remove salt python API terminal output from ceph-boostrap commands
Currently whenever ceph-bootrap runs a salt command through the salt python API, the salt python API prints to the terminal some messages in some situations. Example:

```
No minions matched the target. No command was sent, no jid was assigned.
No minions matched the target. No command was sent, no jid was assigned.
No minions matched the target. No command was sent, no jid was assigned.
```

This is a problem in the salt python API, that should output those messages to the logger instead of directly to stdout, but since we can't change that part here, ceph-bootstrap should wrap the salt python API calls in way that captures the stdout outputs avoid them to be propagated to the user stdout.

We have an example of how that can be achieved here: https://github.com/SUSE/DeepSea/blob/master/cli/common.py#L19-L45

rjfd:
Fixes: #10

### DIFF
--- a/ceph_bootstrap/salt_utils.py
+++ b/ceph_bootstrap/salt_utils.py
@@ -1,3 +1,4 @@
+import contextlib
 import logging
 import os
 import yaml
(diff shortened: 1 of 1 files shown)


ricardoasmarques (approved):
lgtm

PR #51, opened 2020-01-29
cli: require root to run ceph-bootstrap commands

### LINKED ISSUE #18: Root privileges required
If the user does not has root privileges, then `ceph-bootstrap` should return immediately.

See: https://github.com/SUSE/DeepSea/blob/master/cli/common.py#L48-L56

rjfd:
Fixes: #18


smithfarm:
oops, it would appear that Travis does not run the tool as root

ricardoasmarques (approved):
lgtm

PR #62, opened 2020-02-03
qa/ceph_health_test: wait more verbosely

smithfarm:
This commit improves the wait loop in ceph_health_test. First,
there was a bug which caused the loop to run only once, even
when it was supposed to run 5 times.

Second, the loop was not verbose enough and was only checking
the status once per minutes. When running interactively, this
can try the user's patience.


rjfd (approved):
lgtm

PR #68, opened 2020-02-07
ceph_bootstrap/deploy: check and load ceph-salt formula before deployment

### LINKED ISSUE #65: Ensure ceph-salt-formula is loaded by the salt-master before deploy
When running `ceph-bootstrap deploy` we need to make sure that the ceph-salt-formula state files are already loaded by the salt-master, otherwise the deployment will fail.

We can check if ceph-salt formula is loaded by running the following command: `salt \* state.sls_exists ceph-salt`. 
We should also always sync any state, modules or pillar changes before starting the deployment using the following command: `salt \* saltutil.sync_all`

rjfd:
Fixes: #65

### DIFF
--- a/ceph_bootstrap/deploy.py
+++ b/ceph_bootstrap/deploy.py
@@ -1116,7 +1116,47 @@ class CephSaltExecutor:
 
         self.executor = CephSaltExecutorThread(self.controller)
 
+    @staticmethod
+    def check_ceph_salt_formula():
+        # verify that deployment can run
+        PP.println("Checking if ceph-salt formula is available...")
+        result = SaltClient.local_cmd('ceph-salt:member', 'state.sls_exists', ['ceph-salt'],
+                                      tgt_type='grain')
+        if not all(result.values()):
+            PP.println("salt-master will be restarted to load ceph-salt formula")
+            logger.info('restarting salt-master service')
+            result = SaltClient.caller_cmd('service.restart', ['salt-master'])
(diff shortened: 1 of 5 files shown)


ricardoasmarques (approved):
lgtm

ceph_bootstrap/deploy.py
```diff
@@ -1116,7 +1116,47 @@ def __init__(self, interactive):
...
+    @staticmethod
+    def check_ceph_salt_formula():
+        # verify that deployment can run
+        PP.println("Checking if ceph-salt formula is available...")
+        result = SaltClient.local_cmd('ceph-salt:member', 'state.sls_exists', ['ceph-salt'],
+                                      tgt_type='grain')
+        if not all(result.values()):
+            PP.println("salt-master will be restarted to load ceph-salt formula")
+            logger.info('restarting salt-master service')
+            result = SaltClient.caller_cmd('service.restart', ['ceph-salt'])
```
ricardoasmarques: Shouldn't be `salt-master` instead of `ceph-salt`?

ricardoasmarques (changes requested):
Wrong service name?

ceph_bootstrap/deploy.py
rjfd: good catch. Pushed the fix

ricardoasmarques (approved):
lgtm

PR #85, opened 2020-02-19
ceph-salt-formula: eliminate implicit dependency on which

smithfarm:
"which" is not a shell builtin; it is distributed as a separate RPM.

The same functionality (testing whether a given executable is available
in the system) can be obtained using the "type" shell builtin.

Please note that this commit is not introducing a bashism:
`sh -c 'type foo'` should work regardless of which shell `/bin/sh` happens
to be pointing to.

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_orch.py
+++ b/ceph-salt-formula/salt/_modules/ceph_orch.py
@@ -2,7 +2,7 @@
 import json
 
 def configured():
-    ret = __salt__['cmd.run_all']("which ceph")
+    ret = __salt__['cmd.run_all']("sh -c 'type ceph'")
     if ret['retcode'] != 0:
         return False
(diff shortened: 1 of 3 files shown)


smithfarm:
This is a chicken-and-egg problem. The ceph-bootstrap tool needs to have a way to determine whether the "ceph" command is available on any of the minions - if it is, then we can assume that the environment is not pristine - at least some work towards deploying Ceph has been done, and maybe the cluster is already deployed.

The current approach using "which" is not good, because "which" might not be installed on the system. (In upstream ceph, we changed "which" to "type" everywhere already long ago to avoid introducing a dependency on "which".) But the alternative requires bash, and someone might object that we have no guarantee that "bash" is installed on the system, either.

Instead of `bash -c "builtin type ceph"` we could use `rpm -q ceph-common`, but this has the same problem - how do we know `rpm` is available in the system?

ricardoasmarques (approved):
lgtm

PR #88, opened 2020-02-19
Migrate ceph-bootstrap-qa to sesdev (Part 1)

smithfarm:
Part 2 is https://github.com/SUSE/sesdev/pull/112

Rationale:

qa/health-ok.sh is a generic script for smoke-testing Ceph deployments.
It doesn't need to carry any ceph-bootstrap-specific tests.

It makes more sense to have it in sesdev, so sesdev can use it to test
*any* of the Ceph deployments is creates (not just those that use
ceph-bootstrap).


ricardoasmarques (approved):
lgtm

PR #99, opened 2020-02-24
ceph-osd: work around osd deployment "false positive"

smithfarm:
Due to an upstream issue in the Orchestrator, OSD deployment succeeds
without actually deploying any OSDs. Add a "sleep" to work around this
and enable OSD deployment to succeed.

Works-Around: https://tracker.ceph.com/issues/44270

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/ceph-osd.sls
+++ b/ceph-salt-formula/salt/ceph-salt/ceph-osd.sls
@@ -2,6 +2,16 @@
 
 {{ macros.begin_stage('Deployment of Ceph OSDs') }}
 
+{{ macros.begin_step('Spinning wheels for 60 seconds to work around https://tracker.ceph.com/issues/44270') }}
+
+spin wheels for 60 seconds before creating any osds:
+  cmd.run:
+    - name: |
+        sleep 60
+    - failhard: True
+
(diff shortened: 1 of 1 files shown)


ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
```diff
@@ -1,5 +1,15 @@
 {% import 'macros.yml' as macros %}
 
+{% set dg_list = pillar['ceph-salt'].get('storage', {'drive_groups': []}).get('drive_groups', []) %}
+
+{% if dg_list | length == 1 %}
+osd crush chooseleaf for single-node cluster:
+  cmd.run:
+    - name: |
+        echo -en "[global]\n        osd crush chooseleaf type = 0\n" > /root/ceph.conf
```
sebastian-philipp: why not call `ceph config set ` instead?

sebastian-philipp (changes requested):

sebastian-philipp:
Adding workaround is ok with me, but I'd still prefer to have a proper / real solution.

ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
smithfarm: explained in #100

sebastian-philipp (approved):

smithfarm:
Closing in favor of #109 which works around the issue in a more intelligent way.

PR #105, opened 2020-02-27
ceph-mon: automatically set chooseleaf type if needed

### LINKED ISSUE #100: Provide a mechanism for setting "ceph.conf" configuration values so they will be in effect during "cephadm bootstrap" phase
Nowadays, Ceph has a MON store for cluster configuration, but certain options still have to be set in `ceph.conf` before the cluster is bootstrapped.

One example is `osd crush chooseleaf type = 0`. If this is not provided on the `cephadm bootstrap` command line via the `-c` option, the initial CRUSH map created by `cephadm bootstrap` will have the failure domain set to "host" and there is no easy way to change that.

It's possible that there are other options like this one, which must be set via `cephadm bootstrap -c` in order to properly take effect.

Therefore, I am proposing that ceph-salt provide a mechanism for setting these options.

---

UPDATE: I found another situation where this is needed (and I think it's quite likely that there are more):

If someone needs to run `cephadm bootstrap` with MGR debugging turned up, the only way to do this is via `cephadm bootstrap -c`.

smithfarm:
Fixes: https://github.com/ceph/ceph-salt/issues/100

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -1,5 +1,15 @@
 {% import 'macros.yml' as macros %}
 
+{% set dg_list = pillar['ceph-salt'].get('storage', {'drive_groups': []}).get('drive_groups', []) %}
+
+{% if dg_list | length == 1 %}
+osd crush chooseleaf for single-node cluster:
+  cmd.run:
+    - name: |
+        echo -en "[global]\n        osd crush chooseleaf type = 0\n" > /root/ceph.conf
+    - failhard: True
+{% endif %}
+
 {{ macros.begin_stage('Ceph bootstrap') }}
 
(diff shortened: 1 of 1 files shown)


smithfarm:
We should probably think about this before merging it: on the one hand, it **does** fix the single-node deployment case. On the other hand, someone might be surprised that chooseleaf gets set to "osd" when deploying only one OSD node - especially if they intend to deploy additional OSD nodes later.

That's why I suggested to implement a more generic feature enabling the user to specify settings for the bootstrap ceph.conf - something like:

```
ceph-salt config /BootstrapCephConf/Global "osd crush chooseleaf type = 0"
```

smithfarm:
Right now, this is the only option for deploying a single-node cluster (other than letting it be deployed with the wrong CRUSH map and then changing the CRUSH map post-deployment).

ricardoasmarques (approved):
Should be improved by https://github.com/ceph/ceph-salt/issues/100, but for now LGTM

Issue #84, opened 2020-02-19
No bootstrap minion configured if there is no node has both MON and MGR


bk201:
In issue #42 a minion is tagged as bootstrapping role only when it has both MON and MGR roles, if this is a requirement, should we document it (maybe in sesdev or downstream doc) or add a check before deploying?

I tested sesdev with the following command:

```
sesdev create octopus --roles="[admin], [mon], [mgr], [storage]" dev
```

The deployment failed with:
```
    admin: ++ ceph-bootstrap config ls
    admin: o- / ......................................................................................................................... [...]
    admin:   o- Cluster ................................................................................................................. [...]
    admin:   | o- Minions ........................................................................................................ [Minions: 4]
    admin:   | | o- admin.dev.com .................................................................................................. [no roles]
    admin:   | | o- node1.dev.com ....................................................................................................... [mon]
    admin:   | | o- node2.dev.com ....................................................................................................... [mgr]
    admin:   | | o- node3.dev.com .................................................................................................. [no roles]
    admin:   | o- Roles ......................................................................... [Bootstrap minion: None, Minions w/ roles: 2]
    admin:   |   o- Mgr .......................................................................................................... [Minions: 1]
    admin:   |   | o- node2.dev.com .......................................................................................... [no other roles]
    admin:   |   o- Mon .......................................................................................................... [Minions: 1]
    admin:   |     o- node1.dev.com .......................................................................................... [no other roles]
    admin:   o- Containers .............................................................................................................. [...]
    admin:   | o- Images ................................................................................................................ [...]
    admin:   |   o- ceph ..................................................................... [docker.io/ceph/daemon-base:latest-master-devel]
    admin:   o- Deployment .............................................................................................................. [...]
    admin:   | o- Bootstrap ......................................................................................................... [enabled]
    admin:   | o- Dashboard ............................................................................................................. [...]
    admin:   | | o- password ............................................................................................................ [***]
    admin:   | | o- username .......................................................................................................... [admin]
    admin:   | o- Mgr ............................................................................................................... [enabled]
    admin:   | o- Mon ............................................................................................................... [enabled]
    admin:   | o- OSD ............................................................................................................... [enabled]
    admin:   o- SSH ............................................................................................................ [Key Pair set]
    admin:   | o- Private_Key ............................................................... [a4:b9:b1:3e:e7:a2:ba:fe:c6:d6:e5:82:e6:99:d3:24]
[...]

PR #112, opened 2020-03-03
Add "status" command

### LINKED ISSUE #7: Implement `status` command
The `ceph-bootstrap status` command should preform some validations to guarantee that ceph-bootstrap can work correctly in the installed system, and show the status of those checks, and if something fails, suggest possible actions to users.

List of validations:
- Salt
    - check if salt-master package is installed, or just that salt-master is available (it might be installed from source)
    - check if we can communicate with the salt-master, execute operations, etc...
- ceph-salt-formula
    - check if package is installed (it should be if ceph-bootstrap was installed through the RPM)
    - check if state files are present
    - check if salt pillar is correctly configured by looking at `top.sls` file
        - if not, it should suggest to run `ceph-bootstrap init` (Issue #8) that should take care of making sure
          the pillar gets correctly configured.
- Config
  - check that at least one minion is both 'mgr' and 'mon'

ricardoasmarques:
`ceph-salt status` command will check if Salt is correctly configured and if the current configuration is valid.
Additionally, `ceph-salt deploy` command will also check if current configuration is valid.

**Example:**
Screenshot from 2020-03-03 23-23-58

Fixes: https://github.com/ceph/ceph-salt/issues/7

### DIFF
--- a/ceph_salt/__init__.py
+++ b/ceph_salt/__init__.py
@@ -7,7 +7,7 @@ import time
 import click
 import pkg_resources
 
-from .config_shell import run_config_cmdline, run_config_shell
+from .config_shell import run_config_cmdline, run_config_shell, run_status
 from .exceptions import CephSaltException
 from .terminal_utils import check_root_privileges, PrettyPrinter as PP
 from .deploy import CephSaltExecutor
@@ -84,6 +84,19 @@ def config_shell(config_args):
     else:
         if not run_config_shell():
             sys.exit(1)
+
+
+@cli.command(name='status')
+@click.option('-n', '--no-color', is_flag=True, default=False,
+              help='Ouput without colors')
+def status(no_color):
+    """
+    Check ceph-salt status
+    """
+    if no_color:
+        PP.disable_colors()
+    if not run_status():
+        sys.exit(1)
 
 
 @cli.command(name='deploy')
--- a/ceph_salt/config_shell.py
+++ b/ceph_salt/config_shell.py
@@ -12,6 +12,7 @@ from .core import CephNodeManager, SshKeyManager
 from .exceptions import CephSaltException, PillarFileNotPureYaml
 from .salt_utils import PillarManager
 from .terminal_utils import PrettyPrinter as PP
+from .validate.config import validate_config
 from .validate.salt_master import check_salt_master_status, CephSaltPillarNotConfigured
 
 
@@ -812,6 +813,18 @@ base:
     - ceph-salt
 """)
     return False
+
+
+def run_status():
+    if not check_config_prerequesites():
+        return False
+    error_msg = validate_config()
+    if error_msg:
+        logger.info(error_msg)
+        PP.pl_red(error_msg)
+        return False
+    PP.pl_green("OK")
+    return True
 
 
 def run_config_shell():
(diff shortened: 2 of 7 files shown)


bk201:
Test with sesdev:
```
sesdev create octopus --roles="[admin], [mon], [mgr], [storage]" --ceph-salt-repo https://github.com/ricardoasmarques/ceph-salt.git --ceph-salt-branch wip-status --stop-before-ceph-salt-deploy dev2
```

The `ceph-salt status` works as description when there is no node acts as Mgr and Mon
```
admin:~ # ceph-salt status
At least one minion must be both 'Mgr' and 'Mon'
```

But the deploy command can't reach the check:
```
admin:~ # ceph-salt -ldebug deploy --non-interactive
```

ricardoasmarques:
> ...
> But the deploy command can't reach the check:
> 
> ```
> admin:~ # ceph-salt -ldebug deploy --non-interactive
> Checking if ceph-salt formula is available...
> salt-master will be restarted to load ceph-salt formula
> Could not find ceph-salt formula. Please check if ceph-salt-formula package is installed
> ```

@bk201 the check you are facing happens before the config validation, so I think this is not related to this PR.

Check if all minions are returning "True" when executing:

[1]
```
salt -G 'ceph-salt:member' state.sls_exists ceph-salt
```

If no, check if `salt-minion` service is running in all minions:

```
systemctl status salt-minion
```

Try to restart `salt-master` service manually, and check [1] again:

```
systemctl restart salt-master
```

smithfarm:
@ricardoasmarques With `ceph-salt deploy` there is a `--non-interactive` switch which, among other things, prevents ceph-salt from emitting VT-style escape sequences that, in an interactive setting, do nice things like display colors, but elsewhere (like when being run by an automated tool like sesdev and teuthology) have the effect of adding lots of noise - in the form of raw escape sequences - to the log.

Does this new command `ceph-salt status` have a similar option for turning off the escape sequences?

bk201:
> 
> @bk201 the check you are facing happens before the config validation, so I think this is not related to this PR.

Thanks! I checked again and It's actually the undefined variable `bootstrap_minion` in pillar (as the minion log in https://github.com/ceph/ceph-salt/issues/84) makes minions reply error when running `salt -G 'ceph-salt:member' state.sls_exists ceph-salt`. So it can't reach the check here yet to print useful message: https://github.com/ceph/ceph-salt/pull/112/files#diff-d5595865bb360e37be437f3b1446834cR1268-R1273

ricardoasmarques:
> ...
> Does this new command `ceph-salt status` have a similar option for turning off the escape sequences?

@smithfarm I've added a `-n`/`--no-color` option to the `status` command

smithfarm (approved):

ricardoasmarques:
> ...
> 
> Thanks! I checked again and It's actually the undefined variable `bootstrap_minion` in pillar (as the minion log in #84) makes minions reply error when running `salt -G 'ceph-salt:member' state.sls_exists ceph-salt`. So it can't reach the check here yet to print useful message: https://github.com/ceph/ceph-salt/pull/112/files#diff-d5595865bb360e37be437f3b1446834cR1268-R1273

@bk201  Good catch! Should be fixed now, thanks.

bk201 (approved):
Thanks! The new changes work!

PR #113, opened 2020-03-04
Allow users to configure custom registries

bk201:
**Update**: Now supports V2 registry config.

User can configure custom registries as cache of public registries or
pull images from insecure registries.

Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>

## About V2 registry config

V2 config allows configuring image pulling behaviors related to registries :
- Specify alternative location of an image.
  Say users want to do an air-gap deployment, but the image `docker.io/ceph/daemon-base:latest-master-devel` is not reachable from the local LAN. User can mirror the image to a local registry:
  ```
  # using skopeo as an example, in practice user might need to load a tarball
  # skopeo copy --dest-tls-verify=false docker://docker.io/ceph/daemon-base:latest-master-devel docker://192.168.2.106:5000/docker.io/ceph/daemon-base:latest-master-devel
  ```
  
  User then specify the local registry as an alternative location in the config:
  ```
  # cat /etc/containers/registries.conf
  [[registry]]
  prefix = "docker.io"
  location = "192.168.2.106:5000/docker.io"
  insecure = true
  ```
  
  The User then pull the image as usual, but because the config exists, it will directly pull the image from `192.168.2.106:5000/docker.io/ceph/daemon-base:latest-master-devel`:
  ```
  # podman pull docker.io/ceph/daemon-base:latest-master-devel  
  ```

- Use an insecure local registry.
  With the following config:
  ```
  [[registry]]
  location = "192.168.2.106:5000"
  insecure = true
  ```
  
  Users can pull images from `192.168.2.106:5000` even it's not SSL-enabled.

## How to test

Use sesdev to prepare an env and stop before deploy.

### [Bring up a docker registry](https://docs.docker.com/registry/#basic-commands) and push an image into it.
e.g. sync latest octopus image to 192.168.2.106:5000
```
skopeo copy --dest-tls-verify=false docker://registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph docker://192.168.2.106:5000/registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph
```

### Add registries to pillar:

```
ceph-salt config /containers/registries add prefix=registry.opensuse.org location=192.168.2.106:5000/registry.opensuse.org insecure=true
```

### Specify an image from the insecure registry

This step can be skipped. Because by default we already use `registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph` as image path. And because we configured the registries config, podman will pulls image from `172.17.0.1:5000/registry.opensuse.org`.

```
ceph-salt config /containers/images/ceph set registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph
```

### Deploy
```
ceph-salt -ldebug deploy --non-interactive
```

## The example config after states are applied

```
master:/srv/salt # cat /etc/containers/registries.conf 
# This file is managed by ceph-salt.
# For more information on this configuration file, see containers-registries.conf(5)

# An array of host[:port] registries to try when pulling an unqualified image, in order
unqualified-search-registries = ["docker.io"]

[[registry]]

prefix = "quay.io"

location = "192.168.2.106:5000/quay.io"

insecure = true

```

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
+++ b/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
@@ -1,0 +1,20 @@
+{% set registries = pillar['ceph-salt'].get('container', {}).get('registries', []) -%}
+# {% include "ceph-salt/files/managed-header.txt.j2" ignore missing %}
+# For more information on this configuration file, see containers-registries.conf(5)
+
+# An array of host[:port] registries to try when pulling an unqualified image, in order
+unqualified-search-registries = ["docker.io"]
+
+{% for reg in registries %}
+[[registry]]
+{% if reg.prefix is defined %}
+prefix = "{{ reg.prefix }}"
+{% endif %}
+location = "{{ reg.location }}"
+{% if reg.insecure is defined %}
+insecure = {{ '{}'.format(reg.insecure) | lower }}
+{% endif %}
+{% if reg.blocked is defined %}
+blocked = {{ '{}'.format(reg.blocked) | lower }}
+{% endif %}
+{% endfor %}
--- a/ceph_salt/config_shell.py
+++ b/ceph_salt/config_shell.py
@@ -1,4 +1,5 @@
 # pylint: disable=arguments-differ
+import itertools
 import logging
 import fnmatch
 import json
@@ -11,6 +12,7 @@ from configshell_fb.shell import locatedExpr
 
 from .core import CephNodeManager, SshKeyManager, CephNode
 from .exceptions import CephSaltException, MinionDoesNotExistInConfiguration, PillarFileNotPureYaml
+from .params_helper import BooleanStringValidator, BooleanStringTransformer
 from .salt_utils import GrainsManager, PillarManager, SaltClient, CephOrch
 from .terminal_utils import PrettyPrinter as PP
 from .validate.config import validate_config
@@ -624,9 +626,22 @@ class FlagOptionNode(OptionNode):
         PP.pl_green('Disabled.')
 
 
+class KeyValueNode(configshell.ConfigNode):
+    def __init__(self, key, value, parent):
+        configshell.ConfigNode.__init__(self, key, parent)
+        self._value = str(value)
+
+    def summary(self):
+        return self._value, None
+
+
 class ListElementNode(configshell.ConfigNode):
-    def __init__(self, value, parent):
+    def __init__(self, value, parent, child_data=None):
         configshell.ConfigNode.__init__(self, value, parent)
+        if child_data is not None:
+            if isinstance(child_data, dict):
+                for k, v in child_data.items():
+                    KeyValueNode(k, v, self)
 
 
 class ListOptionNode(OptionNode):
@@ -661,6 +676,97 @@ class ListOptionNode(OptionNode):
             PP.pl_green('Value removed.')
         else:
             PP.pl_red('Value not found.')
+
+
+class ListDictOptionNode(OptionNode):
+    def __init__(self, option_name, option_dict, parent):
+        super(ListDictOptionNode, self).__init__(option_name, option_dict, parent)
+        value_list, _ = self._find_value()
+        self.value = list(value_list)
+        self._add_children(self.value)
+
+    def _add_children(self, items):
+        for idx, item in enumerate(items):
+            ListElementNode(str(idx), self, item)
+
+    def _remove_all_children(self):
+        for idx in range(len(self.value)):
+            self.remove_child(self.get_child(str(idx)))
+
+    def _list_commands(self):
+        return ['add', 'remove']
+
+    def summary(self):
+        value_list, _ = self._find_value()
+        return str(len(value_list)) if value_list else 'empty', None
+
+    def _precheck_params(self, spec, kwargs):
+        # check required parameters
+        all_params = set(spec.keys())
+        required_params = {k for k, v in spec.items() if v.get('required', False)}
+        assert required_params  # at least one parameter is required
+        missing_params = required_params - set(kwargs.keys())
+        if missing_params:
+            raise Exception('Required parameter(s) `{}` are missing.'.format(
+                ', '.join(missing_params)))
+        # check if there are any unknown parameters
+        unknown_params = set(kwargs.keys()) - all_params
+        if unknown_params:
+            raise Exception('Unknown parameter(s): `{}`.'.format(', '.join(unknown_params)))
+
+    def _format_params(self, kwargs):
+        """Validate and format input parameters."""
+        spec = self.option_dict.get('params_spec')
+        self._precheck_params(spec, kwargs)
+        value = {}
+        for k, v in kwargs.items():
+            # validate the input value
+            validator = spec[k].get('validator')
+            if validator and not validator.validate(v):
+                raise Exception("Invalid value {} for {}".format(v, k))
+            # transform to native type from str
+            transformer = spec[k].get('transformer')
+            value[k] = transformer.transform(v) if transformer else v
+        return value
+
+    def ui_command_add(self, **kwargs):
+        new_item = self._format_params(kwargs)
+        if new_item not in self.value:
+            ListElementNode(str(len(self.value)), self, new_item)
+            self.value.append(new_item)
+            self.option_dict['handler'].save(self.value)
+            PP.pl_green('Value added.')
+        else:
+            PP.pl_red('Value already exists.')
+
+    def ui_command_remove(self, **kwargs):
+        remove_item = self._format_params(kwargs)
+
+        # do not include items that have specified KVs
+        def __match(item):
+            for search_k, search_v in remove_item.items():
+                if item.get(search_k) != search_v:
+                    break
+            else:
+                return True
+            return False
+        new_value = list(itertools.filterfalse(__match, self.value))
+
+        if len(new_value) != len(self.value):
+            self._remove_all_children()
+            self.value = new_value
+            self._add_children(self.value)
+            self.option_dict['handler'].save(self.value)
+            PP.pl_green('Value removed.')
+        else:
+            PP.pl_red('Value not found.')
+
+    def ui_command_reset(self):
+        '''
+        Resets option value to the default
+        '''
+        self._remove_all_children()
+        super(ListDictOptionNode, self).ui_command_reset()
 
 
 class ConfElementNode(configshell.ConfigNode):
@@ -869,6 +975,8 @@ def _generate_option_node(option_name, option_dict, parent):
         FlagOptionNode(option_name, option_dict, parent)
     elif option_dict.get('type', None) == 'list':
         ListOptionNode(option_name, option_dict, parent)
+    elif option_dict.get('type', None) == 'list_dict':
+        ListDictOptionNode(option_name, option_dict, parent)
     elif option_dict.get('type', None) == 'conf':
         ConfOptionNode(option_name, option_dict, parent)
     elif option_dict.get('type', None) == 'minions':
--- a/ceph_salt/config_shell.py
+++ b/ceph_salt/config_shell.py
@@ -11,7 +11,12 @@ import configshell_fb as configshell
 from configshell_fb.shell import locatedExpr
 
 from .core import CephNodeManager, SshKeyManager, CephNode
-from .exceptions import CephSaltException, MinionDoesNotExistInConfiguration, PillarFileNotPureYaml
+from .exceptions import (
+    CephSaltException,
+    MinionDoesNotExistInConfiguration,
+    PillarFileNotPureYaml,
+    ParamsException
+)
 from .params_helper import BooleanStringValidator, BooleanStringTransformer
 from .salt_utils import GrainsManager, PillarManager, SaltClient, CephOrch
 from .terminal_utils import PrettyPrinter as PP
@@ -352,6 +357,33 @@ CEPH_SALT_OPTIONS = {
                     },
                 }
             },
+            'registries': {
+                'type': 'list_dict',
+                'default': [],
+                'help': '''
+                        List of custom registries in v2 format.
+                        =======================================
+
+                        Add by specifying B{location}, B{prefix}, and B{insecure}. e.g.,
+
+                          add location=172.17.0.1:5000/docker.io prefix=docker.io insecure=true
+                        ''',
+                'params_spec': {
+                    'location': {
+                        'required': True
+                    },
+                    'prefix': {},
+                    'insecure': {
+                        'validator': BooleanStringValidator,
+                        'transformer': BooleanStringTransformer
+                    },
+                    'blocked': {
+                        'validator': BooleanStringValidator,
+                        'transformer': BooleanStringTransformer
+                    }
+                },
+                'handler': PillarHandler('ceph-salt:container:registries')
+            }
         }
     },
     'system_update': {
@@ -707,12 +739,13 @@ class ListDictOptionNode(OptionNode):
         assert required_params  # at least one parameter is required
         missing_params = required_params - set(kwargs.keys())
         if missing_params:
-            raise Exception('Required parameter(s) `{}` are missing.'.format(
+            raise ParamsException('Required parameter(s) are missing: {}.'.format(
                 ', '.join(missing_params)))
         # check if there are any unknown parameters
         unknown_params = set(kwargs.keys()) - all_params
         if unknown_params:
-            raise Exception('Unknown parameter(s): `{}`.'.format(', '.join(unknown_params)))
+            raise ParamsException(
+                'Unknown parameter(s): {}.'.format(', '.join(unknown_params)))
 
     def _format_params(self, kwargs):
         """Validate and format input parameters."""
@@ -723,7 +756,7 @@ class ListDictOptionNode(OptionNode):
             # validate the input value
             validator = spec[k].get('validator')
             if validator and not validator.validate(v):
-                raise Exception("Invalid value {} for {}".format(v, k))
+                raise ParamsException("Invalid value for parameter {}: {}.".format(k, v))
             # transform to native type from str
             transformer = spec[k].get('transformer')
             value[k] = transformer.transform(v) if transformer else v
@@ -735,9 +768,9 @@ class ListDictOptionNode(OptionNode):
             ListElementNode(str(len(self.value)), self, new_item)
             self.value.append(new_item)
             self.option_dict['handler'].save(self.value)
-            PP.pl_green('Value added.')
+            PP.pl_green('Item added.')
         else:
-            PP.pl_red('Value already exists.')
+            PP.pl_red('Item already exists.')
 
     def ui_command_remove(self, **kwargs):
         remove_item = self._format_params(kwargs)
@@ -752,14 +785,15 @@ class ListDictOptionNode(OptionNode):
             return False
         new_value = list(itertools.filterfalse(__match, self.value))
 
-        if len(new_value) != len(self.value):
+        remove_count = len(self.value) - len(new_value)
+        if remove_count > 0:
             self._remove_all_children()
             self.value = new_value
             self._add_children(self.value)
             self.option_dict['handler'].save(self.value)
-            PP.pl_green('Value removed.')
+            PP.pl_green('{} item(s) removed.'.format(remove_count))
         else:
-            PP.pl_red('Value not found.')
+            PP.pl_red('Item not found.')
 
     def ui_command_reset(self):
         '''
--- a/ceph-salt-formula/salt/ceph-salt/container.sls
+++ b/ceph-salt-formula/salt/ceph-salt/container.sls
@@ -1,0 +1,20 @@
+{% import 'macros.yml' as macros %}
+
+{{ macros.begin_stage('Container environment') }}
+
+{{ macros.begin_step('Configure registries') }}
+
+/etc/containers/registries.conf:
+  file.managed:
+    - source:
+        - salt://ceph-salt/files/registries.conf.j2
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: '0644'
+    - backup: minion
+    - failhard: True
+
+{{ macros.end_step('Configure registries') }}
+
+{{ macros.end_stage('Container environment') }}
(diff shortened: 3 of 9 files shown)


ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
```diff
@@ -0,0 +1,29 @@
...
+registries = [
+{% for registry in insecure_registries %}
+  "{{ registry }}",
+{% endfor %}
+]
+{% else %}
+registries = []
+{% endif %}
+
+# Blocked Registries, blocks the `docker daemon` from pulling from the blocked registry.  If you specify
```
smithfarm: does ceph-salt use the docker daemon at all? I thought it used `podman`...

smithfarm:
What is the need that this feature is addressing? It's already possible to point ceph-salt at the latest upstream master container image.

Are you building your own container images locally?

ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
bk201: podman uses the same config file for registries by default.

bk201:
> What is the need that this feature is addressing? It's already possible to point ceph-salt at the latest upstream master container image.
> 
> Are you building your own container images locally?

I need this feature for:
- Speedup the testing of deployment. Every node pulls image from a local registry, not from the Internet.
- Patch my WIP codes into images for testing.

Another possible use scenario is air gap installation. Provisioning a local registry is one method to distribute images. Although in that case generating a self-sign cert and distributing the certs to nodes might be better.

ricardoasmarques:
I'm not 100% sure that this feature should be part of `ceph-salt`.

I see `ceph-salt` as a tool that uses a registry, but setting up a registry seems a bit out of scope to me.

ATM this feature is mainly useful for development environment so I wonder if we can simply add instructions to README on how to set up local/insecure registries that can then be used on `ceph-salt config /Containers/Images/ceph set ...`, instead of adding support for automatically setup those registries.

I might be missing something, thoughts?

smithfarm:
@bk201 It seems like this feature is only for developers, so maybe sesdev would be a better place for it?

bk201:
Thanks for the feedback.

> I'm not 100% sure that this feature should be part of `ceph-salt`.
> 
> I see `ceph-salt` as a tool that uses a registry, but setting up a registry seems a bit out of scope to me.
> 
> ATM this feature is mainly useful for development environment so I wonder if we can simply add instructions to README on how to set up local/insecure registries that can then be used on `ceph-salt config /Containers/Images/ceph set ...`, instead of adding support for automatically setup those registries.
> 
> I might be missing something, thoughts?

The formula doesn't set up the registry. It just adds the registries to the exception list for podman.

> @bk201 It seems like this feature is only for developers, so maybe sesdev would be a better place for it?

Agree. Close this and do it in the sesdev.

toabctl:
This is not only for developers. It's super useful for eg. airgaped installations, registry pull through caches, ...
The idea here is, that you configure the clients (podman on all nodes) to use another registry or use multiple registries (in case one is unavailable). With that, you could setup a local registry (or a pull through cache) and that would reduce the needed bandwith a lot. Eg. having a cluster with 200 nodes would require to download the ceph image 200 times for every update.

ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
```diff
@@ -0,0 +1,29 @@
+# For more information on this configuration file, see containers-registries.conf(5).
+#
+# Registries to search for images that are not fully-qualified.
+# i.e. foobar.com/my_image:latest vs my_image:latest
+[registries.search]
+registries = ["docker.io"]
+
+{% set insecure_registries = pillar['ceph-salt'].get('container', {}).get('insecure_registries', []) %}
```
toabctl: Could we model the data structure in a way where it works also with version 2 file format? With version 2, you would want to create something like:
```
unqualified-search-registries = ["docker.io"]

[[registry]]
prefix = "registry.suse.de"
#location = "registry.suse.de"
location = "10.86.4.8"
insecure = true
#blocked = false
```

So maybe having a list of dicts with registry name and insecure flag?

smithfarm:
@toabctl If it's superuseful (and I'm not disputing that), why implement the feature only for Salt users? Why not have it in cephadm itself?

toabctl:
> @toabctl If it's superuseful (and I'm not disputing that), why implement the feature only for Salt users? Why not have it in cephadm itself?

Fine for me. But imo definitely not a sesdev thing. But you could argue that chrony and friends could also be configured by cephadm .

sebastian-philipp:
> @toabctl If it's superuseful (and I'm not disputing that), why implement the feature only for Salt users? Why not have it in cephadm itself?

Changing the system configuration wasn't really in scope for cephadm till now. (Well, with the exception of [prepare-host](https://github.com/ceph/ceph/blob/master/src/cephadm/cephadm#L2829) which is not used by us.)

ricardoasmarques:
I'm trying to test this PR without success.

My `/etc/containers/registries.conf` is configured, but `podman` still refuses to connect due to HTTP vs HTTPS issue.

Any idea what's wrong here?

```
node3:~ # cat /etc/containers/registries.conf 
# For more information on this configuration file, see containers-registries.conf(5).
#
# Registries to search for images that are not fully-qualified.
# i.e. foobar.com/my_image:latest vs my_image:latest
[registries.search]
registries = ["docker.io"]


# Registries that do not use TLS when pulling images or uses self-signed
# certificates.
[registries.insecure]

registries = [

  "192.168.1.102:500",

]


# Blocked Registries, blocks the `docker daemon` from pulling from the blocked registry.  If you specify
# "*", then the docker daemon will only be allowed to pull from registries listed above in the search
# registries.  Blocked Registries is deprecated because other container runtimes and tools will not use it.
# It is recommended that you use the trust policy file /etc/containers/policy.json to control which
# registries you want to allow users to pull and push from.  policy.json gives greater flexibility, and
# supports all container runtimes and tools including the docker daemon, cri-o, buildah ...
[registries.block]
registries = []


node3:~ # podman pull 192.168.1.102:5000/ceph/daemon-base:latest-master-devel
Trying to pull 192.168.1.102:5000/ceph/daemon-base:latest-master-devel...ERRO[0000] Error pulling image ref //192.168.1.102:5000/ceph/daemon-base:latest-master-devel: Error initializing source docker://192.168.1.102:5000/ceph/daemon-base:latest-master-devel: pinging docker registry returned: Get https://192.168.1.102:5000/v2/: http: server gave HTTP response to HTTPS client 
Failed
Error: error pulling image "192.168.1.102:5000/ceph/daemon-base:latest-master-devel": unable to pull 192.168.1.102:5000/ceph/daemon-base:latest-master-devel: unable to pull image: Error initializing source docker://192.168.1.102:5000/ceph/daemon-base:latest-master-devel: pinging docker registry returned: Get https://192.168.1.102:5000/v2/: http: server gave HTTP response to HTTPS client
``'

**Comment by ricardoasmarques** [comment]
> Any idea what's wrong here?

My bad. I've missed one zero when setting the insecure repository
(should be `add 192.168.1.102:5000` instead of `add 192.168.1.102:500`)

**Review by ricardoasmarques — APPROVED** [review]
lgtm

File: ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2 (hunk 2 above)
Comment (bk201) [reply ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2]: > Could we model the data structure in a way where it works also with version 2 file format? 

Thanks for the feedback. For now, I'll keep that simple. This is just a short term solution, as I mentioned in previous comments, I think the better approach for AIRGAP solution is to

* Setup a local registry
* Put it behind a reverse proxy with SSL on
* Setup authentication 
* Accept the self-gen certs on all minions
* Login to the registry on all minions

**Review by sebastian-philipp — APPROVED** [review]

**Comment by toabctl** [comment]
So I asked the caasp ppl about support for version2 of container-registry.conf. And V2 is supported . So I would prefer to use v2. According to caasp, that's also the only way to support air-gapped installations currently.

File: ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2 (hunk 2 above)
Comment (toabctl) [reply ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2]: This is just a different configuration format for the client side. And V2 can do more things than V1. That's why I would like to see V2 used.

File: ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2 [hunk 3]
```diff
@@ -0,0 +1,18 @@
+{% set registries = pillar['ceph-salt'].get('container', {}).get('registries', []) -%}
+# System Registry Configuration File (VERSION 2)
```
toabctl: - Do we usually have a header to mention that the file is managed by salt?
- nit: can we have as first line `# For more information on this configuration file, see containers-registries.conf(5).` ?

ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
```diff
@@ -0,0 +1,18 @@
...
+
+# An array of host[:port] registries to try when pulling an unqualified image, in order
+unqualified-search-registries = ["docker.io"]
+
+{% for registry in registries -%}
+[[registry]]
+{%- set reg = registry | load_json %}
+prefix = "{{ reg.prefix }}"
```
toabctl, line 11: `prefix` is optional (see `man containers-registries.conf`). So it might not be defined.

ceph_salt/config_shell.py
```diff
@@ -315,6 +315,12 @@ def possible_values(self):
                     },
                 }
             },
+            'Registries': {
+                'type': 'list',
+                'default': [],
+                'help': "List of custom registries",
```
toabctl: maybe `List of custom registries in V2 format` ?

toabctl: Is it possible to also validate that the list elements need to be `dict`?

ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
bk201: Fixed.

ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
bk201: Fixed.  Add a header on template based files

ceph_salt/config_shell.py
bk201: > maybe `List of custom registries in V2 format` ?

Fixed.

> Is it possible to also validate that the list elements need to be `dict`?

Hm..it's actually a JSON string. Let's keep it as it is refine it later, ideally we should change the CLI to the form
```
ceph-salt config /Containers/Registries add prefix=registry.opensuse.org location=192.168.2.106:5000/registry.opensuse.org insecure=true
```

toabctl (approved):
Thanks!

toabctl:
@ricardoasmarques @sebastian-philipp are you still fine with this change?

toabctl:
@bk201 could you rebase to get a CI run, please?

bk201:
> @bk201 could you rebase to get a CI run, please?

Done. But test failed when installing packages.

ceph_salt/config_shell.py
```diff
@@ -315,6 +315,12 @@ def possible_values(self):
                     },
                 }
             },
+            'Registries': {
+                'type': 'list',
+                'default': [],
+                'help': "List of custom registries in V2 format",
```
sebastian-philipp: how do I specify that this registry should be insecure? Should we add this to the help text?

sebastian-philipp (approved):

ceph_salt/config_shell.py
toabctl: good point. It's a list of dicts with specific key names. I wonder how we could document that in the help text. By example maybe?

toabctl:
@bk201 needs another rebase...

ceph_salt/config_shell.py
bk201: Adding a sample in help text.

https://github.com/ceph/ceph-salt/pull/113/files#diff-462eb277dd732ec3b848c7671527a1e8R351-R360

toabctl (approved):

ceph_salt/config_shell.py
```diff
@@ -345,6 +345,21 @@ def possible_values(self):
...
+                        List of custom registries.
+                        ==========================
+                        Add JSON objects in V2 format. e.g.,
+
+                        '{ "prefix":"docker.io", "location":"172.17.0.1:5000/docker.io",
+                           "insecure":true }'
+
+                        Where B{location} is mandatory.
+                        ''',
+                'handler': PillarHandler('ceph-salt:container:registries')
```
ricardoasmarques, line 385: @bk201 now that we finally dropped the Drive_Groups inline-JSON field from `ceph-salt` config, is there any chance we can avoid introducing a field that is a list of inline-JSON strings again?

I imagine a `RegistryHandler` or `RegistryOption`, with multiple parameters[1],  that provides a command like:

```
add location=172.17.0.1:5000/docker.io prefix=docker.io insecure=true
```

And will write the following to `ceph-salt` pillar:

```
container:
    registries:
      - prefix: docker.io
        location: 172.17.0.1:5000/docker.io
        insecure: true
```

The help command will automatically be improved too:

```
> /Containers/Registries help
...
AVAILABLE COMMANDS
==================
...
- add location [prefix] [insecure] [blocked]
```

Thoughts?

[1] `def ui_command_add(self, location, prefix=None, insecure=None, blocked=None):`

bk201: Let's do this way. I wanted to do this and had no idea how to. 
Since the feature is not so urgent, I'll refactor again. Thanks!

ricardoasmarques: @bk201 yeah, sorry for adding another iteration on this PR. I hope that avoiding JSON strings will have benefits.

Here is a POC for the `add` command part: https://github.com/ricardoasmarques/ceph-salt/commit/99a4d33ab4bca499b306c94643db27fd6ca42f37

Screenshot from 2020-03-31 16-29-56

bk201: Thanks for the example. No matter how I tried, the optional parameters are all treated as first `location` parameter. (See the screenshot above, two `insecure`. The first one is in location parameter).

Turn out it's because we changed the parser before (https://github.com/ceph/ceph-salt/commit/40d39a5e8573f57dfdab6a9fc4bed3cf0a488977#diff-960f2a21fca0ad88209fc4bc2e2892e4R744). Since we don't need to capture `"|{|}` anymore, I will reverse the change back.

bk201: @ricardoasmarques How about this?

`ceph-salt config ls`
```
  o- Containers .............................................................................................................. [...]
  | o- Images ................................................................................................................ [...]
  | | o- ceph ....................................................................................... [quay.io/ceph-ci/ceph:octopus]
  | o- Registries .............................................................................................................. [2]
  |   o- 0 ................................................................................................................... [...]
  |   | o- insecure ......................................................................................................... [True]
  |   | o- location ..................................................................... [192.168.2.106:5000/registry.opensuse.org]
  |   | o- prefix .......................................................................................... [registry.opensuse.org]
  |   o- 1 ................................................................................................................... [...]
  |     o- insecure ......................................................................................................... [True]
  |     o- location ................................................................................... [192.168.2.106:5000/quay.io]
  |     o- prefix ........................................................................................................ [quay.io]

```

And corresponding pillar data:

```
  container:
    images:
      ceph: quay.io/ceph-ci/ceph:octopus
    registries:
    - insecure: true
      location: 192.168.2.106:5000/registry.opensuse.org
      prefix: registry.opensuse.org
    - insecure: true
      location: 192.168.2.106:5000/quay.io
      prefix: quay.io

```

ricardoasmarques: > Since we don't need to capture `"|{|}` anymore, I will reverse the change back.

Yes, that "hack" was introduced to support JSON string. I think that we no longer need to override the parser at all.

ricardoasmarques: > @ricardoasmarques How about this?

looks good. thanks for changing it.

bk201: > > Since we don't need to capture `"|{|}` anymore, I will reverse the change back.
> 
> Yes, that "hack" was introduced to support JSON string. I think that we no longer need to override the parser at all.

```diff
--- a/ceph_salt/config_shell.py
+++ b/ceph_salt/config_shell.py
@@ -1015,7 +1015,7 @@ class CephSaltConfigShell(configshell.ConfigShell):
             '~/.ceph_salt_config_shell')
         # Grammar of the command line
         command = locatedExpr(Word(alphanums + '_'))('command')
-        var = Word(alphanums + ';,=_\+/.<>()~@:-%[]*{}" ')
+        var = Word(alphanums + ';,=_\+/.<>()~@:-%[] ')  # adding '*' ' '
         value = var
         keyword = Word(alphanums + '_\-')
         kparam = locatedExpr(keyword + Suppress('=') + Optional(value, default=''))('kparams*')

```

Removing the hack seems breaking `ConfOptionNode` which bootstrap config is using. I need to keep a space in the parser to support bootstrap config, because it allows `my option1 = 2`  to be entered as positional parameter. But this will make keyword parameters like `prefix=docker.io` also be treated as positional parameters. Any suggestions about how to proceed?

ricardoasmarques: @bk201 good point. Using that `=` in the `ConfOptionNode` was not a good idea.

I'll create a PR to change the way user set config options from:

```
ceph-salt config /Cephadm_Bootstrap/Ceph_Conf/global set osd crush chooseleaf type = 0
```
to 
```
ceph-salt config /Cephadm_Bootstrap/Ceph_Conf/global set "osd crush chooseleaf type" 0
```

After that, we will be able to drop that `space` from the parser.

ricardoasmarques: @bk201 I've now submitted the PR that should fix your issue, please take a look: https://github.com/ceph/ceph-salt/pull/162

ceph_salt/config_shell.py
```diff
@@ -345,6 +345,21 @@ def possible_values(self):
                     },
                 }
             },
+            'Registries': {
```
ricardoasmarques: nit: Since https://github.com/ceph/ceph-salt/pull/166 we are now using lowercase on config nodes, can you please rename it?

bk201: Thanks for reminding me. Fixed too.

ceph_salt/config_shell.py
bk201: Thanks for the groundwork. Instead of creating a special node for the registry. I create a new node type that can let the user specify a list of dictionaries, any need in the future can reuse se it too. Please take a look on acffea8

ceph_salt/config_shell.py
```diff
@@ -663,6 +705,97 @@ def ui_command_remove(self, value):
...
+                return True
+            return False
+        new_value = list(itertools.filterfalse(__match, self.value))
+
+        if len(new_value) != len(self.value):
+            self._remove_all_children()
+            self.value = new_value
+            self._add_children(self.value)
+            self.option_dict['handler'].save(self.value)
+            PP.pl_green('Value removed.')
```
ricardoasmarques: nit: For consistency, and better usability, can we show the number of removed items? Something like https://github.com/ceph/ceph-salt/blob/master/ceph_salt/config_shell.py#L837-L840

ricardoasmarques (changes requested):
Executing `remove` without `location` parameter results in a Traceback. Can we fail more gracefully here?
```
/containers/registries> remove
Traceback (most recent call last):
  File "/usr/bin/ceph-salt", line 11, in <module>
    load_entry_point('ceph-salt==15.2.0', 'console_scripts', 'ceph-salt')()
  File "/usr/lib/python3.6/site-packages/ceph_salt/__init__.py", line 55, in ceph_salt_main
    cli(prog_name='ceph-salt')
  File "/usr/lib64/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib64/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib64/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib64/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib64/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/ceph_salt/__init__.py", line 90, in config_shell
    if not run_config_shell():
  File "/usr/lib/python3.6/site-packages/ceph_salt/config_shell.py", line 1117, in run_config_shell
    shell.run_interactive()
  File "/usr/lib/python3.6/site-packages/configshell_fb/shell.py", line 900, in run_interactive
    self._cli_loop()
  File "/usr/lib/python3.6/site-packages/configshell_fb/shell.py", line 729, in _cli_loop
    self.run_cmdline(cmdline)
  File "/usr/lib/python3.6/site-packages/configshell_fb/shell.py", line 843, in run_cmdline
    self._execute_command(path, command, pparams, kparams)
  File "/usr/lib/python3.6/site-packages/configshell_fb/shell.py", line 818, in _execute_command
    result = target.execute_command(command, pparams, kparams)
  File "/usr/lib/python3.6/site-packages/configshell_fb/node.py", line 1406, in execute_command
    return method(*pparams, **kparams)
  File "/usr/lib/python3.6/site-packages/ceph_salt/config_shell.py", line 770, in ui_command_remove
    remove_item = self._format_params(kwargs)
  File "/usr/lib/python3.6/site-packages/ceph_salt/config_shell.py", line 747, in _format_params
    self._precheck_params(spec, kwargs)
  File "/usr/lib/python3.6/site-packages/ceph_salt/config_shell.py", line 738, in _precheck_params
    ', '.join(missing_params)))
Exception: Required parameter(s) `location` are missing.
```

ceph_salt/config_shell.py
bk201: Hmm.. how about renaming the term to `items`? add item to a list, remove items from list. `Value` is a bit weird for dictionaries.

bk201: Fixed, thanks!

bk201:
> Executing `remove` without `location` parameter results in a Traceback. Can we fail more gracefully here?

Fixed by raising appropriate exceptions. Thanks!

ricardoasmarques (approved):
LGTM

PR #114, opened 2020-03-04
ceph-salt: deploy additional MGRs in same way as additional MONs

smithfarm:
"ceph orch apply mgr" is asynchronous:

```
admin:~ # ceph orch apply mgr 3 node3:10.20.174.203
Scheduled mgr update...
```

But ceph-salt needs to add new mgrs in a synchronous way, to avoid
timing issues.


jschmid1 (approved):
cephadm should actually try to avoid timing issues itself..

lgtm for the moment, though

smithfarm:
Unfortunately, it doesn't work. The only way I have found to avoid https://bugzilla.suse.com/show_bug.cgi?id=1165492 is to have ceph-salt deploy only one MON and one MGR, and then add the additional MONs and MGRs manually afterwards.

Update: even that way only worked for me once, and I was unable to reproduce it.

Issue #118, opened 2020-03-09
Adapt ceph-salt to latest {mon,mgr,osd} deployment UI changes


smithfarm:
https://github.com/ceph/ceph/pull/33792 was recently merged, which once again changed the command-line UI for creating MONs, MGRs, and OSDs.

ceph-salt will need to be adapted again.

bk201:
For OSD, we need to adapt to this change: https://github.com/ceph/ceph/pull/33922
Either using `orch apply osd`/`orch daemon add osd`, or `orch apply -i` with a big config.

sebastian-philipp:
I think, `orch apply -i` is broken right now. See https://github.com/ceph/ceph/pull/34001

PR #119, opened 2020-03-10
Fix error when deploying additional mgrs

bk201:
mgr deployment fails because of PlacementSpec change in
https://github.com/ceph/ceph/pull/33706.

Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>


bk201:
This doesn't address all aspects in https://github.com/ceph/ceph-salt/issues/118.
But it's an immediate fix to error when running with upstream images.

ricardoasmarques (approved):
lgtm

Issue #120, opened 2020-03-11
Do not install ceph-common on "Admin" nodes


ricardoasmarques:
Instead of installing `ceph-common` in "Admin" nodes, `ceph-salt` should simply create a `ceph='cephadm shell -- ceph'` alias (e.g. /etc/profile.d/ceph.sh)

**Depends on https://github.com/ceph/ceph/pull/33829**

ricardoasmarques:
Maybe we can do the same for crushtool: `cephadm shell -- crushtool --help`

smithfarm:
As Denis Polom reported in Rocket.Chat, this approach is not robust, because if the command line references a file or path that is on the host, but not in the container, the command will fail.

For example:

```
# cephadm shell -- crushtool -d file.bin -o file.txt
INFO:cephadm:Using recent ceph image registry.suse.de/suse/sle-15-sp2/update/products/ses7/milestones/containers/ses/7/ceph/ceph:latest
crushtool: error reading 'file.bin': can't open file.bin: (2) No such file or directory
```

This could be surprising if:

* the file is present on the host and
* the user issues the crushtool on the host but it gets converted into `cephadm shell -- crushtool $@` by an alias added (more or less silently) by ceph-salt to `/etc/bash.d`

ricardoasmarques:
Good point, I think this is a valid argument to not create an alias, and force the user to type `cephadm shell` explicitly.

ricardoasmarques:
Using `cephadm shell -- ceph` instead of `ceph` has usability issues especially when referencing files via `--mount` option.

For this reason, we will continue installing `ceph-common` for now.

PR #121, opened 2020-03-11
Add "Admin" role

### LINKED ISSUE #71: Admin role
ATM, `ceph-common` and `/etc/ceph/ceph.client.admin.keyring` are installed on all "mons".

Instead, we should have a dedicated "Admin" role, and only install ceph tools on minions that have that role.

Links
https://github.com/ceph/ceph/pull/33793

ricardoasmarques:
Instead of installing "ceph-tools [1]" on all "Mgr" nodes, we will now install those tools on"Admin" nodes.

Note that "Bootstrap" minion must be an "Admin" node.

[1] ceph.conf, keyring, and ceph-common[2]
[2] ceph-common will be replaced by an alias on https://github.com/ceph/ceph-salt/issues/120

Fixes: https://github.com/ceph/ceph-salt/issues/71

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/ceph-admin.sls
+++ b/ceph-salt-formula/salt/ceph-salt/ceph-admin.sls
@@ -1,0 +1,20 @@
+{% import 'macros.yml' as macros %}
+
+{% if pillar['ceph-salt']['minions']['admin'] | length > 1 %}
+
+{{ macros.begin_stage('Copying ceph.conf and keyring to other Admin nodes') }}
+
+copy ceph.conf and keyring to other admin nodes:
+  cmd.run:
+    - name: |
+{%- for minion in pillar['ceph-salt']['minions']['admin'] %}
+{%- if minion != grains['host'] %}
+        scp -o "StrictHostKeyChecking=no" /etc/ceph/ceph.conf root@{{ minion }}:/etc/ceph/
+        scp -o "StrictHostKeyChecking=no" /etc/ceph/ceph.client.admin.keyring root@{{ minion }}:/etc/ceph/
+{%- endif %}
+{%- endfor %}
+    - failhard: True
+
+{{ macros.end_stage('Copying ceph.conf and keyring to other Admin nodes') }}
+
+{% endif %}
--- a/ceph-salt-formula/salt/ceph-salt/cephtools.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephtools.sls
@@ -8,7 +8,7 @@ install cephadm:
   pkg.installed:
     - pkgs:
         - cephadm
-{% if 'mon' in grains['ceph-salt']['roles'] %}
+{% if 'admin' in grains['ceph-salt']['roles'] %}
         - ceph-common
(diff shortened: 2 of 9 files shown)


ceph-salt-formula/salt/ceph-salt/cephtools.sls
```diff
@@ -8,7 +8,7 @@ install cephadm:
   pkg.installed:
     - pkgs:
         - cephadm
-{% if 'mon' in grains['ceph-salt']['roles'] %}
+{% if 'admin' in grains['ceph-salt']['roles'] or grains['id'] == pillar['ceph-salt']['bootstrap_minion'] %}
         - ceph-common
```
smithfarm, line 12: I thought the idea was to *not* install ceph-common?

smithfarm: ah - that's going to be a separate PR. Got it.

ceph-salt-formula/salt/ceph-salt/ceph-admin.sls
```diff
@@ -0,0 +1,20 @@
...
+    - name: |
+{%- for minion in pillar['ceph-salt']['minions']['admin'] %}
+{%- if minion != grains['host'] %}
+        scp -o "StrictHostKeyChecking=no" /etc/ceph/ceph.conf root@{{ minion }}:/etc/ceph/
+        scp -o "StrictHostKeyChecking=no" /etc/ceph/ceph.client.admin.keyring root@{{ minion }}:/etc/ceph/
+{%- endif %}
+{%- endfor %}
+    - failhard: True
+
+{{ macros.end_stage('Deployment of Admin nodes') }}
```
smithfarm: Here I'm concerned what will happen if there is no "separate" Admin node to copy files to. It seems like ceph-salt will produce output like this:

```
[STEP] [BEGIN] Deployment of Admin nodes
[STEP] [END] Deployment of Admin nodes
```

Won't the user find this confusing if he did not define any Admin role?

Also, it seems that, in the "no Admin roles defined" case, the `cmd.run` state will be invoked with an empty `name`.

In other places I saw this code pattern:

```
{%- for minion in pillar['ceph-salt']['minions']['admin'] %}
{%- if minion != grains['host'] %}

copy ceph.conf and keyring to Admin node {{ minion }}:
  cmd.run:
    - name: |
        scp -o "StrictHostKeyChecking=no" /etc/ceph/ceph.conf root@{{ minion }}:/etc/ceph/
        scp -o "StrictHostKeyChecking=no" /etc/ceph/ceph.client.admin.keyring root@{{ minion }}:/etc/ceph/
    - failhard: True
{%- endif %}
{%- endfor %}
```

It seems like writing it this way would at least eliminate the possibility of an empty `cmd.run` invocation.

ricardoasmarques: Should be fixed now, please review again.

smithfarm (approved):

bk201 (approved):
works!

PR #129, opened 2020-03-17
Support bootstrap ceph config

### LINKED ISSUE #100: Provide a mechanism for setting "ceph.conf" configuration values so they will be in effect during "cephadm bootstrap" phase

ricardoasmarques:
It's now possible to specify `ceph.conf` options that will be used for boostrap (`cephadm bootstrap -c ...`)

Usage example:

```
master:~ # ceph-salt config /Deployment/Bootstrap_Ceph_Conf/Global set "osd crush chooseleaf type = 0"
Value set.
master:~ # ceph-salt config /Deployment/Bootstrap_Ceph_Conf ls
o- Bootstrap_Ceph_Conf ....................................................................................................... [...]
  o- Global .................................................................................................................. [...]
    o- osd crush chooseleaf type ............................................................................................... [0]

```

Fixes: https://github.com/ceph/ceph-salt/issues/100

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -1,18 +1,22 @@
 {% import 'macros.yml' as macros %}
 
-{% set dg_list = pillar['ceph-salt'].get('storage', {'drive_groups': []}).get('drive_groups', []) %}
-
-{% if dg_list | length == 1 %}
-osd crush chooseleaf for single-node cluster:
-  cmd.run:
-    - name: |
-        echo -en "[global]\n        osd crush chooseleaf type = 0\n" > /root/ceph.conf
-    - failhard: True
-{% endif %}
-
 {{ macros.begin_stage('Ceph bootstrap') }}
 
 {% if grains['id'] == pillar['ceph-salt']['bootstrap_minion'] %}
+
+{% set bootstrap_ceph_conf = pillar['ceph-salt'].get('bootstrap_ceph_conf', {}) %}
+
+create bootstrap ceph conf:
+  cmd.run:
+    - name: |
+        echo -en "" > /tmp/bootstrap-ceph.conf
+{% for section, settings in bootstrap_ceph_conf.items() %}
+        echo -en "[{{ section }}]\n" >> /tmp/bootstrap-ceph.conf
+{% for setting, value in settings.items() %}
+        echo -en "        {{ setting }} = {{ value }}\n" >> /tmp/bootstrap-ceph.conf
+{% endfor %}
+{% endfor %}
+    - failhard: True
 
 {{ macros.begin_step('Wait for other minions') }}
 wait for other minions:
@@ -30,9 +34,7 @@ run cephadm bootstrap:
     - name: |
         CEPHADM_IMAGE={{ pillar['ceph-salt']['container']['images']['ceph'] }} \
         cephadm --verbose bootstrap --mon-ip {{ grains['fqdn_ip4'][0] }} \
-{%- if dg_list | length == 1 -%}
-                --config /root/ceph.conf \
-{%- endif %}
+                --config /tmp/bootstrap-ceph.conf \
                 --initial-dashboard-user {{ dashboard_username }} \
                 --output-keyring /etc/ceph/ceph.client.admin.keyring \
                 --output-config /etc/ceph/ceph.conf \
--- a/ceph_salt/config_shell.py
+++ b/ceph_salt/config_shell.py
@@ -369,6 +369,12 @@ CEPH_SALT_OPTIONS = {
                 'handler': PillarHandler('ceph-salt:deploy:osd'),
                 'default': False
             },
+            'Bootstrap_Ceph_Conf': {
+                'type': 'conf',
+                'help': 'Bootstrap Ceph configuration',
+                'default': [],
+                'handler': PillarHandler('ceph-salt:bootstrap_ceph_conf')
+            },
             'Dashboard': {
                 'type': 'group',
                 'help': 'Dashboard settings',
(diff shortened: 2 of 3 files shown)


ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
```diff
@@ -21,6 +11,17 @@ osd crush chooseleaf for single-node cluster:
...
 
+{% set global_configs = pillar['ceph-salt'].get('bootstrap_ceph_conf', {}).get('global', {}) %}
+
+create bootstrap ceph conf:
+  cmd.run:
+    - name: |
+        echo -en "[global]\n" > /tmp/bootstrap-ceph.conf
+{% for k, v in global_configs.items() %}
+        echo -en "        {{ k }} = {{ v }}\n" >> /tmp/bootstrap-ceph.conf
+{% endfor %}
```
sebastian-philipp: not every item should end up in the `[global]` section. can we simply pass-through the config?

ceph_salt/config_shell.py
```diff
@@ -369,6 +369,18 @@ def possible_values(self):
                 'handler': PillarHandler('ceph-salt:deploy:osd'),
                 'default': False
             },
+            'Bootstrap_Ceph_Conf': {
+                'type': 'group',
+                'help': 'Bootstrap Ceph configuration',
+                'options': {
+                    'Global': {
+                        'type': 'dict',
```
sebastian-philipp: can we change this to a simple multi-line string?

ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
ricardoasmarques: I've now changed it to not create the `[global]` section if no global configs are defined.

ceph_salt/config_shell.py
ricardoasmarques: I've started by using the already existing `list` type, but then I've created the `dict` type where the option name and value are separated giving us some usability advantages:

1) Option name is displayed on the left, and value on the right (more consistency with other `ceph-salt` options)

2) It's possible to rewrite a setting value by simply setting it again (in `list` type we have to delete the old entry and create the new entry to avoid duplicated option)

Screenshot from 2020-03-18 11-21-31

Are you suggesting to change it from `dict` type to `list` type because you are anticipating any issue with `dict` or any advantage in using `list`?

sebastian-philipp: I'm just suggesting that we shouldn't be to clever here. I just want to avoid a maintenance burden, if someone wants us to add support for the `[mon]` section. And then later someone wants us to add support for the `[osd]` section, and then later someone wants us to add support for the `[client]` section and then someone wants us to add support for the `[mon.a]` section. And the someone wants us to add support for the `[client.x.y]` section. And then someone wants us to add support for the `[cephfs-shell]` section. And so on.

sebastian-philipp: my idea, if users can simply add a multiline string

```
[global]
...
[mon]
....
[mon.a]
...
[client.x]
...
```
we're done.

ricardoasmarques: Now I understand what you mean by `multiline string`, I was confusing it with `list of strings`.

I wonder if it's possible to implement `multiline string` input in configshell.

How can the user add/edit that value?

ricardoasmarques: I think we can support all types of `sections` by allowing the user to create sections, e.g.:

```
ceph-salt /Deployment/Bootstrap_Ceph_Conf add global
ceph-salt /Deployment/Bootstrap_Ceph_Conf/global set "<key> = <val>"
```

smithfarm:
Is this still a "WIP"? Or is it ready for testing already?

ricardoasmarques:
@smithfarm @sebastian-philipp I've now pushed a new implementation where it's possible to create config sections, so please review again.

Usage example:

Screenshot from 2020-03-24 17-42-16

smithfarm:
I'm testing this with a simplified version of https://github.com/SUSE/sesdev/pull/183

```
    master: ++ ceph-salt config /Deployment/Bootstrap_Ceph_Conf add global
    master: Section added.
    master: ++ ceph-salt config /Deployment/Bootstrap_Ceph_Conf/global set osd crush chooseleaf type = 0
    master: Setting set.
```

```
    master:   | o- Bootstrap .........................................................................................................
 [enabled]
    master:   | o- Bootstrap_Ceph_Conf ................................................................................................... [...]
    master:   | | o- global .............................................................................................................. [...]
    master:   | |   o- osd crush chooseleaf type ........................................................................................... [0]
```

```
    master: ++ ceph-salt export --pretty
    master: {
    master:     "bootstrap_ceph_conf": {
    master:         "global": {
    master:             "osd crush chooseleaf type": "0"
    master:         }
    master:     },
```

All good.

And it also seems to work as expected, both with and without `--single-node`.

ceph_salt/config_shell.py
```diff
@@ -655,6 +661,124 @@ def ui_command_remove(self, value):
...
+        child = None
+        if setting in self.value:
+            child = self.get_child(setting)
+        self.value[setting] = value
+        self.option_dict['handler'].save(self.value)
+        if child:
+            child.value = value
+        else:
+            ConfElementNode(setting, value, self)
+        PP.pl_green('Setting set.')
```
smithfarm: "Setting set" sounds odd in English. Could it be rephrased?

I would prefer: "Setting added." This goes nicely with "Setting removed."

smithfarm: ~Now that I think about it, "osd crush chooseleaf type = 0" is not a setting that can have a value.~

Oops, now I see it is a setting with a value, and I guess the value can even be changed once it's set?

smithfarm: Feel free to ignore these comments. "Setting set" does sound a bit weird, but I'm prepared to live with it.

Maybe "Parameter set." / "Parameter removed." / "Parameter changed." would be nicer, instead of "Setting".

ricardoasmarques:
@smithfarm thanks for the feedback. I've now renamed "Setting" to Parameter".

smithfarm (approved):

toabctl:
Could you add useful commit messages please? At least for me (as a ceph newbie), that would help to understand things (and a future version of me using `git blame` will be more than happy to understand why things changed).

ricardoasmarques:
> Could you add useful commit messages please? At least for me (as a ceph newbie), that would help to understand things (and a future version of me using `git blame` will be more than happy to understand why things changed).

@toabctl I've updated the commit message. Please take another look.

toabctl (approved):

toabctl:
@ricardoasmarques thanks!

Issue #135, opened 2020-03-24
cephadm: CEPHADM_STRAY_DAEMON: bootstrap mgr not managed by ceph


sebastian-philipp:
<pre>
# ceph -s
  cluster:
    id:     6cf878a8-6dbb-11ea-81f8-fa163e09adda
    health: HEALTH_WARN
            1 stray daemons(s) not managed by cephadm
 
  services:
    mon: 1 daemons, quorum host1 (age 12m)
    mgr: host1.rpcqxx(active, since 11m), standbys: host4.xgjlhi, host2.lnnfdk
    osd: 12 osds: 12 up (since 6m), 12 in (since 6m)
 
  data:
    pools:   1 pools, 1 pgs
    objects: 0 objects, 0 B
    usage:   12 GiB used, 108 GiB / 120 GiB avail
    pgs:     1 active+clean
 
# ceph health detail
HEALTH_WARN 1 stray daemons(s) not managed by cephadm
[WRN] CEPHADM_STRAY_DAEMON: 1 stray daemons(s) not managed by cephadm
    stray daemon mgr.host1.rpcqxx on host host1 not managed by cephadm
</pre>

<pre>
# ceph orch ps
NAME              HOST   STATUS         REFRESHED  AGE  VERSION  IMAGE NAME        IMAGE ID      CONTAINER ID  
mgr.host2.lnnfdk  host2  running (19m)  4m ago     19m  15.2.0   ceph/ceph:latest  4947e2bbb809  4257d130525e  
mgr.host4.xgjlhi  host4  running (19m)  4m ago     19m  15.2.0   ceph/ceph:latest  4947e2bbb809  9d9a9f6965ad  
mon.host1         host1  running (21m)  4m ago     -    15.2.0   ceph/ceph:latest  4947e2bbb809  614e444780ff  
osd.0             host1  running (18m)  4m ago     18m  15.2.0   ceph/ceph:latest  4947e2bbb809  1a363bd5f838  
osd.1             host1  running (18m)  4m ago     18m  15.2.0   ceph/ceph:latest  4947e2bbb809  87ba39c1378e  
osd.10            host2  running (15m)  4m ago     15m  15.2.0   ceph/ceph:latest  4947e2bbb809  0fc086350f3b  
osd.11            host2  running (15m)  4m ago     15m  15.2.0   ceph/ceph:latest  4947e2bbb809  9703a260c81f  
[...]

PR #136, opened 2020-03-24
ceph-mgr.sls: do not omit bootstrap MGR from "ceph orch apply mgr"

smithfarm:
With ceph v15.2.0 this code was having the effect of removing the
bootstrap MGR when adding additional MGRs, which of course is not
desired.


ricardoasmarques (approved):
lgtm

ricardoasmarques:
@smithfarm does this PR fixes https://github.com/ceph/ceph-salt/issues/135 ? 

If yes, please update the commit message accordingly, thanks.

smithfarm:
> @smithfarm does this PR fixes #135 ?

I don't know TBH

Update: I would say "no" it doesn't because in #135 the bootstrap manager is still running when ceph-salt finishes, whereas in my case it is completely removed and not running.

PR #137, opened 2020-03-24
cephtools.sls: run "cephadm check-host" on all minions

smithfarm:
Given that cephadm has this functionality, it seems prudent to make use
of it.


ricardoasmarques (approved):
lgtm

PR #143, opened 2020-03-25
time.sls: run "chronyc makestep" before starting chrony.service

### LINKED ISSUE #141: Cluster comes up in "HEALTH_WARN: clock skew detected"
I've been deploying octopus 15.2.0 clusters on SLE-15-SP2 and consistently they are coming up with:

```
HEALTH_WARN
clock skew detected on mon.node
```

The reason for this is ceph-salt starts chronyd on the cluster nodes, but does not wait for them to sync. Typically there is a multi-second (even tens of seconds) skew between the `time_server` cluster node and the external time servers when chronyd first starts. Now, of course, chronyd will eventually sync, but the larger the skew, the longer it takes.

Since this initial health warning has a negative psychological impact on the user (who may not even know what a "clock skew" is), it would be nicer if we avoided the warning by syncing the clocks **before** deploying any MONs.

smithfarm:
By ensuring that chronyd is started and synced on time_server first,
and on the "chrony slaves", second (as opposed to doing it all at once,
in parallel), we prevent the cluster from starting up in "HEALTH_WARN
clock skew detected"

Fixes: https://github.com/ceph/ceph-salt/issues/141

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/init.sls
+++ b/ceph-salt-formula/salt/ceph-salt/init.sls
@@ -6,6 +6,8 @@ include:
     - .software
     - .apparmor
     - .time
+    - .time-server
+    - .time-slave
     - .cephtools
     - .provision-end
 {% if pillar['ceph-salt'].get('deploy', {'bootstrap': True}).get('bootstrap', True) %}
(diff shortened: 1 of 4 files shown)


toabctl (approved):

smithfarm:
this is not so simple

PR #149, opened 2020-03-27
Prevent image download if already downloaded

swiftgist:
Airgapped clusters cannot directly contact registries.  Saving and
loading container images is a valid strategy.  Avoid failing if the
image already exists.

Signed-off-by: Eric Jackson <swiftgist@gmail.com>

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephtools.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephtools.sls
@@ -35,11 +35,14 @@ have cephadm check the host:
 
 {{ macros.end_step('Run "cephadm check-host"') }}
 
+{% set image = pillar['ceph-salt']['container']['images']['ceph'] %}
+
 {{ macros.begin_step('Download ceph container image') }}
 download ceph container image:
   cmd.run:
     - name: |
-        cephadm --image {{ pillar['ceph-salt']['container']['images']['ceph'] }} pull
+        cephadm --image {{ image }} pull
+    - unless: podman image exists {{ image }}
     - failhard: True
 {{ macros.end_step('Download ceph container image') }}
 


toabctl:
I think for airgapped solutions, you should use a local registry (which can be configured with https://github.com/ceph/ceph-salt/pull/113 ). Not sure if this PR is the right approach tbh.

ricardoasmarques:
@swiftgist which failure are you trying to avoid?

I've tried to run the pull command multiple times and I don't get any error:

```
master:~ # cephadm --image registry.suse.de/devel/storage/7.0/containers/ses/7/ceph/ceph pull
INFO:cephadm:Pulling latest registry.suse.de/devel/storage/7.0/containers/ses/7/ceph/ceph...
{
    "ceph_version": "ceph version 15.2.0-29-g274f7bc2e7 (274f7bc2e716d748d124ee9c50451afa2d27ec83) octopus (rc)",
    "image_id": "640486d840f15843abc3f9510050368d96aca2c4d8b318805c1637365cec1c55"
}
master:~ # cephadm --image registry.suse.de/devel/storage/7.0/containers/ses/7/ceph/ceph pull
INFO:cephadm:Pulling latest registry.suse.de/devel/storage/7.0/containers/ses/7/ceph/ceph...
{
    "ceph_version": "ceph version 15.2.0-29-g274f7bc2e7 (274f7bc2e716d748d124ee9c50451afa2d27ec83) octopus (rc)",
    "image_id": "640486d840f15843abc3f9510050368d96aca2c4d8b318805c1637365cec1c55"
} 
```

swiftgist:
For example, configure your DNS such that the hostname doesn't resolve.  Podman doesn't check that the image already exists locally and simply fails.

ml8mr:
> I think for airgapped solutions, you should use a local registry (which can be configured with #113 ). Not sure if this PR is the right approach tbh.

Are you saying airgapped environments _must_ use a local registry? If it's only a recommendation, it seems reasonable to not fail if there isn't a network path to the upstream registry.

toabctl:
> Are you saying airgapped environments _must_ use a local registry? If it's only a recommendation, it seems reasonable to not fail if there isn't a network path to the upstream registry.

@ml8mr What other options beside using a local registry do we have? Where should the container images come from when we eg. want to add another node to the cluster? Or do image updates?

ceph-salt-formula/salt/ceph-salt/cephtools.sls
```diff
@@ -35,11 +35,14 @@ have cephadm check the host:
...
 
+{% set image = pillar['ceph-salt']['container']['images']['ceph'] %}
+
 {{ macros.begin_step('Download ceph container image') }}
 download ceph container image:
   cmd.run:
     - name: |
-        cephadm --image {{ pillar['ceph-salt']['container']['images']['ceph'] }} pull
+        cephadm --image {{ image }} pull
+    - unless: podman image exists {{ image }}
```
toabctl, line 45: This will change the expected behavior of a pull. If you have set the image to eg `registry.opensuse.org/opensuse/ceph/ceph:latest` I would expect that it takes the **latest** image available on the registry. With this code change, it will never check again if there is a newer version of the image available.
Or am I missing something here?

sebastian-philipp: I'd go the kubernetes route and discourage the usage of `:latest` altogether.

ricardoasmarques: Or if we want to change the default `pull` behavior to NOT pull the latest image if we already have a local one, then maybe this can be implemented in `cephadm pull` directly instead of `ceph-salt`? 

Something like:
```
# cephadm pull 
Local image already exists.
# cephadm pull --force
Pulling image....
``` 

Thoughts?

ml8mr: > This will change the expected behavior of a pull. If you have set the image to eg `registry.opensuse.org/opensuse/ceph/ceph:latest` I would expect that it takes the **latest** image available on the registry. With this code change, it will never check again if there is a newer version of the image available.
> Or am I missing something here?

I think you are right if ":latest" is being used, and if so, you definitely want the images to be updated even if you have already downloaded them. Requiring a "--force" doesn't seem to be intuitive or helpful in this case.

toabctl: > I'd go the kubernetes route and discourage the usage of `:latest` altogether.

`latest` was just an example. It could also be `15` which would always pull the latest minor/patch releases for the major version `15`

sebastian-philipp:
> > Are you saying airgapped environments _must_ use a local registry? If it's only a recommendation, it seems reasonable to not fail if there isn't a network path to the upstream registry.
> 
> @ml8mr What other options beside using a local registry do we have? Where should the container images come from when we eg. want to add another node to the cluster? Or do image updates?

I think so, too. We're dealing with a distributed system across multiple hosts. Not having a central registry sounds wrong to me. Besides, @tserong was able to run cephadm using a local registry.

tserong:
> I think so, too. We're dealing with a distributed system across multiple hosts. Not having a central registry sounds wrong to me. Besides, @tserong was able to run cephadm using a local registry.

Yeah, for my test VMs, I ran `podman run -d -p 5000:5000 --restart=always --name registry registry:2` on my VM host, used `skopeo` to copy some images to it from a real registry, then added "192.168.121.1:5000" to registries.insecure in /etc/containers/registries.conf on each test VM. Having done that I could then use e.g. `cephadm --image 192.168.121.1:5000/devel/storage/7.0/containers/ses/7/ceph/ceph [...]`.

ricardoasmarques:
FYI, https://github.com/ceph/ceph-salt/pull/113 has been merged

ricardoasmarques:
I'm closing this PR in favor of https://github.com/ceph/ceph-salt/pull/113.

If this PR is still needed, feel free to reopen.

PR #150, opened 2020-03-27
ci: Drop .github/ directory with workflows

toabctl:
We use travis and tox so the github workflows are not needed
currently.


smithfarm:
We have the same over at https://github.com/SUSE/sesdev and I thought about deleting this directory, but then it occurred to me:

* what if the GitHub workflows become available again, and in the meantime we have forgotten that we had code in the repo to use them?

In this case, I can easily envision someone "reinventing the wheel", whereas if we keep the code in the repo (it does no harm to carry it), this "reinvention of the wheel" cannot happen because whoever decides to re-implement GitHub workflows will see the directory containing the code.

If we delete the code, people would have to (1) know or suspect that the code used to be present, and (2) search the git history for it, and I'm not confident both of these would happen.

toabctl:
@smithfarm so should I close the PR?

smithfarm:
> @smithfarm so should I close the PR?

I personally don't see a compelling reason to rip out this directory right now. If you prefer, you could keep the PR open - maybe others will weigh in?

ricardoasmarques:
What about keeping those files in the repo, but adding a `.github/workflows/READ_THIS` file that explains we are not using it? Something equivalent to:

https://github.com/SUSE/sesdev/blob/master/.github/READ_THIS

ricardoasmarques:
Closing in favor of https://github.com/ceph/ceph-salt/pull/178

PR #155, opened 2020-03-31
Install private/public keys on admin nodes

ricardoasmarques:
'Mgr' role no longer exists, so private and public keys must be set on 'Admin' nodes instead.


smithfarm (commented):
lgtm

PR #156, opened 2020-04-01
Allow explicit set bootstrap Mon IP

### LINKED ISSUE #64: Allow explicit set Mon IP
ATM, ceph-bootstrap always uses the IP address from the `fqdn_ip4` grain to configure Monitors.

This address can be used by default, but it should be possible to explicitly change it to a different value.

This can be done in the `/Deployment` group, e.g.:
```
 ...
 o- Deployment ............................................................... [...]
  | ...
  | o- Mon ............................................................... [enabled]
  | | o- node1.ceph.com .......................................... [192.169.100.201]
 ...
```

When implementing this we should remove the `CephNodeFqdnResolvesToLoopback` validation when adding a `minion`, and only validate IPs before the deployment.

ricardoasmarques:
This PR will allow the user to explicitly set the "Mon IP" used on `cephadm bootstrap --mon-ip <Mon_IP>` command option, instead of always use the IP address return by `fqdn_ip4` salt grain (which forces the user to edit the `/etc/hosts` file).

Note that, for better usability,  when `/Ceph_Cluster/Roles/Bootstrap` minion is set by the user, `ceph-salt`will automatically set `/Cephadm_Bootstrap/Mon_IP` with the IP address return by `fqdn_ip4` salt grain.

```
master:~ # ceph-salt config /Cephadm_Bootstrap ls
o- Cephadm_Bootstrap .......................................................... [enabled]
  o- Ceph_Conf .................................................................... [...]
  o- Dashboard .................................................................... [...]
  | o- password ................................................................... [***]
  | o- username ................................................................. [admin]
  o- Mon_IP .............................................................. [10.20.75.201]
```

Fixes: https://github.com/ceph/ceph-salt/issues/64

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -33,7 +33,7 @@ run cephadm bootstrap:
   cmd.run:
     - name: |
         CEPHADM_IMAGE={{ pillar['ceph-salt']['container']['images']['ceph'] }} \
-        cephadm --verbose bootstrap --mon-ip {{ grains['fqdn_ip4'][0] }} \
+        cephadm --verbose bootstrap --mon-ip {{ pillar['ceph-salt']['bootstrap_mon_ip'] }} \
                 --config /tmp/bootstrap-ceph.conf \
                 --initial-dashboard-user {{ dashboard_username }} \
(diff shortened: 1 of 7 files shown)


ricardoasmarques:
CC @gekios

gekios:
Oh thanks, looks good :)

sebastian-philipp (approved):

toabctl (approved):

PR #165, opened 2020-04-06
Allow explicit set chrony subnet

### LINKED ISSUE #161: Allow explicit set chrony subnet
ATM  https://github.com/ceph/ceph-salt/blob/master/ceph-salt-formula/salt/ceph-salt/files/chrony.conf.j2#L9 is always assuming a subnet based on `fqdn_ip4` grains.

We cannot assume it, so we will need a new config so the user can explicitly set the subnet to use in chrony.

ricardoasmarques:
Instead of relying on `fqdn_ipv4` grain, it's now possible to explicitly set the time server subnet:

```
o- Time_Server ........................................................................... [enabled]
  o- External_Servers .......................................................................... [1]
  | o- 0.pt.pool.ntp.org ..................................................................... [...]
  o- Server_Hostname .......................................................... [master.octopus.com]
  o- Subnet ........................................................................ [10.20.91.0/24]
```
 
Note that for convenience, `ceph-salt` will automatically set the `Subnet` when `Server_Hostname` is set by the user.

Also, when typing `set <tab>`, `ceph-salt` will sugest/auto-complete possible subnet values.

Fixes: https://github.com/ceph/ceph-salt/issues/161

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/files/chrony.conf.j2
+++ b/ceph-salt-formula/salt/ceph-salt/files/chrony.conf.j2
@@ -7,12 +7,7 @@ pool {{ server }} iburst
 
 local stratum 10
 manual
-{%- set fqdn_ip = grains['fqdn_ip4'] %}
-{%- for subnet in salt['network.subnets']() %}
-{%- if salt['network.ip_in_subnet'](fqdn_ip, subnet) %}
-allow {{ subnet }}
-{%- endif %}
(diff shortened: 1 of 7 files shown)


ricardoasmarques:
cc @jan--f

jan--f:
yup makes sense to me. I'll give it a test run next week. #174 should fix the other side of my issue.

smithfarm (approved):

PR #166, opened 2020-04-06
Use lowecase on config nodes

ricardoasmarques:
Using lower-case on config tree nodes is easier to type during navigation (e.g. `cd /ceph_cluster/minions`), and consistent with other tools like `targetcli` and `gwcli`.


sebastian-philipp:
:100:  as long as the CI passes

sebastian-philipp (approved):

PR #169, opened 2020-04-07
Case insensitive config paths

ricardoasmarques:
With this PR, config paths will be case insensitive.


smithfarm (approved):

ricardoasmarques:
To implement this feature we are forced to override a private `method` from `configshell`.

Since we cannot guarantee this method will not be removed or changed in the future, and this is not a mandatory feature, I prefer to close this PR for now.

Ideally, the support for case-insensitive paths should be implemented in `configshell` itself.

PR #174, opened 2020-04-09
core: avoid 127.0.0.1 as a nodes public_ip

### LINKED ISSUE #160: Don't rely on fqdn_ipv4 grain for mon address
We currently rely on the `fqdn_ipv4` grain to populate a nodes public address (see https://github.com/ceph/ceph-salt/blob/master/ceph_salt/core.py#L41). This is not very reliable in that it can contain `127.0.0.1` but also might return an ip address that is from, say a mgmt network.

We need a different way here. Also how can a user influence this choice?

jan--f:
If fqdn_ipv4 grain contains 127.0.0.1, retrieve ipv4 grain and take the
first non 127. ip address.

Fixes: https://github.com/ceph/ceph-salt/issues/160

Signed-off-by: Jan Fajerski <jfajerski@suse.com>

### DIFF
--- a/ceph_salt/core.py
+++ b/ceph_salt/core.py
@@ -39,7 +39,15 @@ class CephNode:
             self.execution = result[self.minion_id]['execution']
 
         result = GrainsManager.get_grain(self.minion_id, 'fqdn_ip4')
-        self.public_ip = result[self.minion_id][0]
+        public_ip = result[self.minion_id][0]
+        if public_ip == '127.0.0.1':
+            logger.debug('fqdn_ipv4 grain is 127.0.0.1, falling back to ipv4 grain')
+            result = GrainsManager.get_grain(self.minion_id, 'ipv4')
+            for addr in result[self.minion_id]:
+                if addr != '127.0.0.1':
+                    public_ip = addr
+                    break
+        self.public_ip = public_ip
 
     def add_role(self, role):
         self.roles.add(role)


jan--f:
ping @ricardoasmarques @smithfarm

ricardoasmarques (approved):
lgtm

ceph_salt/core.py
```diff
@@ -39,7 +39,15 @@ def _load(self):
...
-        self.public_ip = result[self.minion_id][0]
+        public_ip = result[self.minion_id][0]
+        if public_ip == '127.0.0.1':
+            logger.debug('fqdn_ipv4 grain is 127.0.0.1, falling back to ipv4 grain')
+            result = GrainsManager.get_grain(self.minion_id, 'ipv4')
+            for addr in result[self.minion_id]:
+                if addr != '127.0.0.1':
+                    public_ip = addr
+                    break
+        self.public_ip = public_ip
```
toabctl, line 50: Shouldn't that fail here if we still have `127.0.0.1`?

ricardoasmarques: It will fail latter when you run `ceph-salt status` or `ceph-salt deploy` - https://github.com/ceph/ceph-salt/blob/master/ceph_salt/validate/config.py#L20

jan--f: This scenario (`public_ip` still coming up as `127.0.0.1`) would mean that either the node has no ip besides `127.0.0.1` or that salt is messing up populating the ipv4 grain.

We can probably ignore the first case, since we would not have a functioning salt cluster. For the second case at least a log message would be nice I guess.

ricardoasmarques: https://github.com/ceph/ceph-salt/issues/183

PR #175, opened 2020-04-09
Support adding new hosts after initial deployment

### LINKED ISSUE #147: Support adding new hosts after initial deployment
I have a cluster that was already deployed, but when I try to deploy a new "admin" node, ceph.conf/keyring is not installed and the host is not added to orchestrator:

**How to reproduce**

1) Add new minion, set "admin" role and deploy it
```
# ceph-salt config /ceph_cluster/minions add <new-host-name>
# ceph-salt config /ceph_cluster/roles/admin add <new-host-name>
# ceph-salt deploy <new-host-name>
```

2) Check that, after deployment,  ceph.conf and keyring does not exists on `<new-host-name>:/etc/ceph/`

3) Also, check that the host was not added to orchestrator (regardless of having "admin" role or not)
```
# ceph orch host ls
...
```

ricardoasmarques:
As describes in issue https://github.com/ceph/ceph-salt/issues/147, `ceph-salt deploy <minion_id>` is not working correctly, this PR fixes it.

Fixes: https://github.com/ceph/ceph-salt/issues/147

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -2,7 +2,8 @@
 import json
 
 def get_remote_grain(host, grain):
-    ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no root@{} "
+    ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
+                                  "-i /tmp/ceph-salt-ssh-id_rsa root@{} "
                                   "'salt-call grains.get --out=json --out-indent=-1 {}'".format(
                                       host, grain))
     if ret['retcode'] != 0:
@@ -10,5 +11,6 @@ def get_remote_grain(host, grain):
     return json.loads(ret['stdout'])['local']
 
 def set_remote_grain(host, grain, value):
-    return __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no root@{} "
-                                  "'salt-call grains.set {} {}'".format(host, grain, value))
+    return __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
+                                   "-i /tmp/ceph-salt-ssh-id_rsa root@{} "
+                                   "'salt-call grains.set {} {}'".format(host, grain, value))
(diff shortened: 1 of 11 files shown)


smithfarm (approved):
Looks great to me, and works as advertised

ceph-salt-formula/salt/_modules/ceph_salt.py
```diff
@@ -2,13 +2,15 @@
 import json
 
 def get_remote_grain(host, grain):
-    ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no root@{} "
+    ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
+                                  "-i /tmp/ceph-salt-ssh-id_rsa root@{} "
```
sebastian-philipp, line 6: fyi, mgr/cephadm uses 

https://github.com/ceph/ceph/blob/master/src/pybind/mgr/cephadm/module.py#L63-L69

for its ssh config.

ceph-salt-formula/salt/ceph-salt/cephorch.sls
```diff
@@ -0,0 +1,13 @@
+{% import 'macros.yml' as macros %}
+
+{{ macros.begin_stage('Add host to ceph orchestrator') }}
+
+{% set bootstrap_host = pillar['ceph-salt']['bootstrap_minion'].split('.', 1)[0] %}
+
+add host to ceph orch:
+  cmd.run:
+    - name: |
+        ssh -o "StrictHostKeyChecking=no" -i /tmp/ceph-salt-ssh-id_rsa root@{{ bootstrap_host }} "ceph orch host add {{ grains['host'] }}"
```
sebastian-philipp: Can we do this on an admin host instead? What happens if this minion is no longer part of the cluster?

smithfarm: Also, even when this is done on an admin host, the command `ceph orch host add` will fail if and when `ceph-common` is no longer installed on the host...

ricardoasmarques: @sebastian-philipp I've pushed a new commit that will not rely on the bootstrap minion to get ceph `config`+`keyring` and to run `ceph orch host add`.

ricardoasmarques: @smithfarm Correct. That is something to be addressed during the implementation of https://github.com/ceph/ceph-salt/issues/120

PR #201, opened 2020-04-27
[WIP] Do not install "ceph-common"

### LINKED ISSUE #120: Do not install ceph-common on "Admin" nodes

ricardoasmarques:
**Blocked by https://github.com/ceph/ceph/pull/34773**

---

`cephadm shell -- ceph ....` should be used instead of `ceph`

Fixes: https://github.com/ceph/ceph-salt/issues/120

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_orch.py
+++ b/ceph-salt-formula/salt/_modules/ceph_orch.py
@@ -2,18 +2,18 @@
 import json
 
 def configured():
-    ret = __salt__['cmd.run_all']("sh -c 'type ceph'")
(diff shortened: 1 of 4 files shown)


ricardoasmarques:
Using `cephadm shell -- ceph` instead of `ceph` has usability issues especially when referencing files via `--mount` option.

For this reason, we will continue installing `ceph-common` for now.

PR #206, opened 2020-04-29
time: support time server not managed by ceph-salt

smithfarm:
After: #211

With this patch, users can specify an arbitrary time_server_host. If the
time_server_host is a minion managed by ceph-salt, it will be configured
as before. If it is *not* a minion managed by ceph-salt, all the minions
will be pointed at it.

---

To Do:

- [x] documentation
- [x] #211 merged
- [x] this PR rebased
- [x] test standard configuration (master pointed at pool.ntp.org), single-node and 4-node
- [x] test "non-standard" configuration (all nodes pointed at pool.ntp.org), 4-node


ceph-salt-formula/salt/ceph-salt/time-sync.sls
```diff
@@ -1,5 +1,6 @@
 {% if pillar['ceph-salt']['time_server'].get('enabled', True) %}
 {% set time_server = pillar['ceph-salt']['time_server']['server_host'] %}
+{% set time_server_is_minion = pillar['ceph-salt']['time_server']['is_minion' %}
```
ricardoasmarques: Missing a `]` here, although I think we can calculate this on the fly, instead of storing it in the pillar.

I've tested the following:

```diff
+{% set time_server_short_name = pillar['ceph-salt']['time_server']['server_host'].split('.', 1)[0] %}
+{% set time_server_is_minion = time_server_short_name in pillar['ceph-salt']['minions']['all'] %}
+
+{% if time_server_is_minion %}
+is minion - yes:
+  test.nop
+{% else %}
+is minion - no:
+  test.nop
+{% endif %}
```

And it works fine:
```
master:~ # salt node1\* state.show_sls ceph-salt | grep 'is minion'
    is minion - yes:

master:~ # ceph-salt config /time_server/server_hostname set some.host.name
Value set.

master:~ # salt node1\* state.show_sls ceph-salt | grep 'is minion'
    is minion - no:
```

ceph-salt-formula/salt/ceph-salt/time-sync.sls
```diff
@@ -1,5 +1,7 @@
 {% if pillar['ceph-salt']['time_server'].get('enabled', True) %}
 {% set time_server = pillar['ceph-salt']['time_server']['server_host'] %}
+{% set time_server_short_name = time_server.split('.', 1)[0] %}
+{% set time_server_is_minion = time_server_short_name in pillar['ceph-salt']['minions']['all'] %}
```
smithfarm: @ricardoasmarques I guess we should wait until #210 is fixed, right? Then we will be able to:

    {%  set time_server_is_minion = time_server in pillar['ceph-salt']['minions']['all'] %}

ricardoasmarques: @smithfarm Yes, here is the PR https://github.com/ceph/ceph-salt/pull/211

gekios:
I ran my tests on top of this PR successfully so lgtm

ceph_salt/validate/config.py
```diff
@@ -18,13 +19,18 @@ def validate_config(host_ls):
...
-    all_minions = PillarManager.get('ceph-salt:minions:all')
+    time_server_enabled = PillarManager.get('ceph-salt:time_server:enabled', True)
+    if time_server_enabled:
+        time_server_host = PillarManager.get('ceph-salt:time_server:server_host')
+        if not time_server_host:
+            return 'No time server host specified in config'
+        time_server_is_minion = time_server_host in all_minions
+        time_server_subnet = PillarManager.get('ceph-salt:time_server:subnet')
+        if time_server_is_minion and not time_server_subnet:
+            return 'No time server subnet specified in config'
```
ricardoasmarques, line 32: Can we also validate that if `time_server_is_minion`, then `time_server_subnet` cannot be specified?

ceph_salt/validate/config.py
```diff
@@ -18,13 +19,18 @@ def validate_config(host_ls):
...
+        external_time_servers = PillarManager.get('ceph-salt:time_server:external_time_servers')
+        if time_server_is_minion and not external_time_servers:
+            return 'No external time servers specified in config'
```
ricardoasmarques, line 37: Can we also validate that if `time_server_is_minion`, then `external_time_servers` cannot be specified?

ricardoasmarques (approved):
lgtm

PR #207, opened 2020-04-29
Wait for admin should fail if any admin failed

### LINKED ISSUE #196: After initial reboot `ceph-salt deploy` says it can't find ceph-salt formula
Seen on SLE 12 SP2 with salt-3000 and ceph-salt-15.2.1+1587155350.g440cd72.

Initial run of `ceph-salt deploy` resulted in reboot of all nodes, with master node saying "Salt master must be rebooted manually", as shown here:

salt-master-must-be-rebooted-manually

After rebooting the master, subsequent runs of `ceph-salt deploy` resulted in:

```
# ceph-salt deploy
```

Of course, ceph-salt forumula *is* installed. The problem is somehow related to two minions not picking up the ceph-salt:execution:provisioned and/or ceph-salt:execution:failed grain correctly (two minions were printing `[salt.loaded.int.module.cmdmod:397 ][INFO    ][1310] Executing command 'ssh -o StrictHostKeyChecking=no -i /tmp/ceph-salt-ssh-id_rsa root@node1 'salt-call grains.get --out=json --out-indent=-1 ceph-salt:execution:provisioned'' in directory '/root'` every five seconds in /var/log/salt/minion).

Eventually a timeout was hit (https://github.com/ceph/ceph-salt/blob/master/ceph-salt-formula/salt/_states/ceph_salt.py#L62) and `ceph-salt deploy` started behaving correctly again.

ricardoasmarques:
The `wait_for_admin_host` state should fail If any `admin` minion failed execution.

Example:
---

1) Both `/system_update/packages` and `/system_update/reboot` are enabled
2) All `minions` rebooted automatically, but `master` must be rebooted manually so the execution fail on `master`
3) Because `master` failed, then `bootstrap minion` will fail with `"One or more minions failed."` (while waiting for all other minions to be provisioned)
4) Since `boostrap minion` failed, all other minions that are waiting an an `admin` node should fail instead of waiting until the timeout is reached.

This PR implements `4)`.

Screenshot from 2020-04-29 15-59-29

Fixes: https://github.com/ceph/ceph-salt/issues/196

### DIFF
--- a/ceph-salt-formula/salt/_states/ceph_orch.py
+++ b/ceph-salt-formula/salt/_states/ceph_orch.py
@@ -16,6 +16,10 @@ def wait_for_admin_host(name, timeout=1800):
         time.sleep(5)
         admin_hosts = __pillar__['ceph-salt']['minions']['admin']
         for admin_host in admin_hosts:
+            failed = __salt__['ceph_salt.get_remote_grain'](admin_host, 'ceph-salt:execution:failed')
+            if failed:
+                ret['comment'] = 'One or more admin minions failed.'
+                return ret
             provisioned = __salt__['ceph_salt.get_remote_grain'](admin_host,
                                                                 'ceph-salt:execution:provisioned')
             if provisioned:


tserong (approved):
LGTM (although, we'll want to document this behaviour somewhere if we didn't already, because seeing "one or more admin minions failed" doesn't exactly make it obvious that what you need to do is reboot the master and run the command again)

PR #211, opened 2020-04-30
Store minion_id in pillar instead of hostname

### LINKED ISSUE #210: Store minion_id in pillar instead of hostname
ATM, we are storing `short_name`s in pillar

```
ceph-salt:
  bootstrap_minion: node1.octopus-reboot.com
...
  minions:
    admin:
    - node1
    - node2
    - node3
    - master
    all:
    - node1
    - node2
    - node3
    - master
```

`ceph-salt` should always store `minion_ids` and only calculate the short name when needed  https://github.com/ceph/ceph-salt/blob/3eb6982e47d9a8b458b72d50c5be898deba24301/ceph_salt/core.py#L22

ricardoasmarques:
With this PR we will always store the `minion_id` in pillar, instead of the `hostname`.

Also, we are now using the `host` grain do get the minion hostname used on `cephadm orch` commands, instead of relying on `minion_id.split('.', 1)[0]` to infer that.

Note that minion ID must resolve to an IP address (see README).

Fixes: https://github.com/ceph/ceph-salt/issues/210

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ This repository contains two components:
 
 # Setup
 
-In order to use `ceph-salt`, you need a working Salt cluster.
+In order to use `ceph-salt`, you need a working Salt cluster and minion IDs
+must be resolvable to IP addresses (e.g. `host <minion_id>`).
 
 Now, install `ceph-salt` on your Salt Master from the openSUSE
 repositories:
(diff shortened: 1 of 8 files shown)


smithfarm (approved):
Tested

PR #214, opened 2020-05-04
Do not modify registries.conf when not needed

toabctl:
When no container/registries salt pillar is set, do not modify
/etc/containers/registries.conf . That way, users can modify the file
manually and it won't be overwritten when not configured within
ceph-salt.

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/container.sls
+++ b/ceph-salt-formula/salt/ceph-salt/container.sls
@@ -4,6 +4,8 @@
 
 {{ macros.begin_step('Configure registries') }}
 
+{% with registries = pillar['ceph-salt'].get('container', {}).get('registries', []) %}
+  {% if registries|length > 0 %}
 /etc/containers/registries.conf:
   file.managed:
     - source:
@@ -14,6 +16,10 @@
     - mode: '0644'
     - backup: minion
     - failhard: True
+    - defaults:
+        registries: {{ registries }}
+  {% endif %}
+{% endwith %}
 
 {{ macros.end_step('Configure registries') }}
 
--- a/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
+++ b/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
@@ -1,4 +1,3 @@
-{% set registries = pillar['ceph-salt'].get('container', {}).get('registries', []) -%}
 # {% include "ceph-salt/files/managed-header.txt.j2" ignore missing %}
 # For more information on this configuration file, see containers-registries.conf(5)
 


toabctl:
See https://bugzilla.suse.com/show_bug.cgi?id=1170915

ceph-salt-formula/salt/ceph-salt/container.sls
```diff
@@ -14,6 +16,8 @@
     - mode: '0644'
     - backup: minion
     - failhard: True
+  {% endif %}
```
bk201, line 21:
```suggestion
    - defaults:
        registries: {{ registries }}
  {% endif %}
```
I got this error when testing:
```[2020-05-04 05:33:47.825560] [master.mini.com ] Finished with failures

Failure in minion: master.mini.com
__id__: /etc/containers/registries.conf
__run_num__: 6
__sls__: ceph-salt.container
changes: {}
comment: 'Unable to manage file: Jinja variable ''registries'' is undefined'
duration: 40.112
name: /etc/containers/registries.conf
result: false
start_time: '07:33:47.759825'
state: file_|-/etc/containers/registries.conf_|-/etc/containers/registries.conf_|-managed
```

The variable `registries` seems not brought into template automatically.

bk201:
Thanks for creating the PR. I had this check in my original PR, but I removed it because of this edge case:

- Adding registries. e.g.: `ceph-salt config /containers/registries add location=192.168.2.106:5000/quay.io prefix=quay.io insecure=true` 
- Then applies salt states, `/etc/containers/registries.conf` is managed.
- Removing registries with reset. e.g. `ceph-salt config /containers/registries reset`
- Apply the sale states again, because there is no registries in the pillar, so the file templating is skipped. (`/etc/containers/registries.conf` still contains one registry)

I couldn't find a good fix to this problem, but I think for now it's OK to skip this case.

ceph-salt-formula/salt/ceph-salt/container.sls
toabctl: Done

toabctl:
> Thanks for creating the PR. I had this check in my original PR, but I removed it because of this edge case:
> 
> * Adding registries. e.g.: `ceph-salt config /containers/registries add location=192.168.2.106:5000/quay.io prefix=quay.io insecure=true`
> * Then applies salt states, `/etc/containers/registries.conf` is managed.
> * Removing registries with reset. e.g. `ceph-salt config /containers/registries reset`
> * Apply the sale states again, because there is no registries in the pillar, so the file templating is skipped. (`/etc/containers/registries.conf` still contains one registry)
> 
> I couldn't find a good fix to this problem, but I think for now it's OK to skip this case.

I would skip that case for now.

bk201 (approved):

toabctl:
@bk201 did you test it? Because I didn't...

bk201:
> @bk201 did you test it? Because I didn't...

Yes. I tested the latest push.

ricardoasmarques:
We should not decide if `/etc/containers/registries.conf` is managed by `ceph-salt` or not, just by looking into the number of configured registries.

Instead, we should make it explicit just like we did in  `/timeserver` [1], so if we don't want `ceph-salt` to manage `/etc/containers/registries.conf` we should be able to do:

```
ceph-salt config /containers/registries disable
```

(enabled by default)

This also addresses https://github.com/ceph/ceph-salt/pull/214#issuecomment-623270213

[1] https://github.com/ceph/ceph-salt/blob/master/ceph-salt-formula/salt/ceph-salt/time-prep.sls#L1

toabctl:
> ...
> This also addresses [#214 (comment)](https://github.com/ceph/ceph-salt/pull/214#issuecomment-623270213)

How does this address the issue described? If you enable the registry handing and then disable it, > you still have the modified `/etc/containers/registries.conf` . This is not different afaiu.
Or how would you "reset" `/etc/containers/registries.conf` ?

ricardoasmarques:
> How does this address the issue described? If you enable the registry handing and then disable it, > you still have the modified `/etc/containers/registries.conf` . This is not different afaiu.
> Or how would you "reset" `/etc/containers/registries.conf` ?

When you `disable` it you are just saying: **"don't touch `/etc/containers/registries.conf` anymore, regardless of whether it was manually configured or not."**

ricardoasmarques:
Closing in favor of https://github.com/ceph/ceph-salt/pull/262

PR #219, opened 2020-05-07
Default values are now persisted in pillar data

### LINKED ISSUE #60: Default values shared between "configshell code" and "ceph-salt-formulas"
ATM, default values are duplicated in "configshell code" and "ceph-salt-formulas".

It would be great if we have a way to share constants/default values between "configshell code", and "ceph-salt-formulas" (e.g., default dashboard username).

### LINKED ISSUE #177: Need a clear definition of `reset` command
It would be great to have a clear definition of `reset` command. Does a reset command:
- reset pillar data to an empty state (by default it deletes the key), or
- reset pillar data to the default value that is defined in ceph-salt

Taking path `/system_update/packages` as an example, enabling it with CLI makes the key `ceph-salt:updates:enabled` set to true in pillar data:
```
$ ceph-salt config /system_update/packages enable
$ cat /srv/pillar/ceph-salt.sls
  ...
  updates:
    enabled: true
    reboot: false
```

Resetting it makes the key be deleted:
```
$ ceph-salt config /system_update/packages reset
$ cat /srv/pillar/ceph-salt.sls
  ...
  updates:
    reboot: false
```

But the definition of the config in code has a default value set to True:

https://github.com/ceph/ceph-salt/blob/c08206ccc86a27e9784113e688dec40e4aa543cd/ceph_salt/config_shell.py#L396-L401

So, deleting the key in pillar data might be not the original intention when the code is written.

---

I ask this question because I was confused when debugging the following issue:

Currently, the default value that coded in the config is mainly used for the "summary" of config nodes, which was displayed inside the bracket at ends of lines when using interactive mode. 

```
master:~ # ceph-salt config /system_update/packages enable
Enabled.
master:~ # ceph-salt config
/system_update/packages> ls
o- packages ..................... ... .......... [enabled]
/system_update/packages> reset
Value reset.
/system_update/packages> ls
o- packages ..................... ... .......... [enabled]  <-- but the pillar data is gone.
/system_update/packages> 
```

### LINKED ISSUE #192: Unable to see dashboard password
For security reasons we are hiding the dashboard password on `ls` command, but we should have an `info` command on dashboard password node that can be used to see the password.

ricardoasmarques:
With this PR, default pillar values will be persisted when `ceph-salt config` is executed for the first time:

```
master:~ # cat /srv/pillar/ceph-salt.sls
ceph-salt: {}

master:~ # ceph-salt config
/> exit

master:~ # cat /srv/pillar/ceph-salt.sls
ceph-salt:
  dashboard:
    username: admin
  updates:
    enabled: true
    reboot: true
```

This PR resolves the following issues:

1) Default values are no longer duplicated in `ceph-salt` and `ceph-salt-formula`, see https://github.com/ceph/ceph-salt/issues/60

2) When `reset` command is used, default value is writen to pillar data (instead of removed), see https://github.com/ceph/ceph-salt/issues/177

3) This new approach will allow us to generate a random dashboard password in `ceph-salt`, in order to fix https://github.com/ceph/ceph-salt/issues/192

Fixes: https://github.com/ceph/ceph-salt/issues/60
Fixes: https://github.com/ceph/ceph-salt/issues/177

### DIFF
--- a/ceph_salt/validate/config.py
+++ b/ceph_salt/validate/config.py
@@ -14,12 +14,25 @@ def validate_config(host_ls):
             return "No bootstrap minion specified in config"
         if bootstrap_minion not in admin_minions:
             return "Bootstrap minion must be 'Admin'"
+        dashboard_username = PillarManager.get('ceph-salt:dashboard:username')
+        if not dashboard_username:
+            return "No dashboard username specified in config"
         bootstrap_mon_ip = PillarManager.get('ceph-salt:bootstrap_mon_ip')
         if not bootstrap_mon_ip:
             return "No bootstrap Mon IP specified in config"
         if bootstrap_mon_ip in ['127.0.0.1', '::1']:
             return 'Mon IP cannot be the loopback interface IP'
-    time_server_enabled = PillarManager.get('ceph-salt:time_server:enabled', True)
+
+    # system_update
+    if not isinstance(PillarManager.get('ceph-salt:updates:enabled'), bool):
+        return "'ceph-salt:updates:enabled' must be of type Boolean"
+    if not isinstance(PillarManager.get('ceph-salt:updates:reboot'), bool):
+        return "'ceph-salt:updates:reboot' must be of type Boolean"
+
+    # time_server
+    time_server_enabled = PillarManager.get('ceph-salt:time_server:enabled')
+    if not isinstance(time_server_enabled, bool):
+        return "'ceph-salt:time_server:enabled' must be of type Boolean"
     if time_server_enabled:
         time_server_host = PillarManager.get('ceph-salt:time_server:server_host')
         if not time_server_host:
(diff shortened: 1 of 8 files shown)


ceph_salt/validate/config.py
```diff
@@ -14,11 +14,21 @@ def validate_config(host_ls):
...
+        if not dashboard_username:
+            return "No dashboard username specified in config"
         bootstrap_mon_ip = PillarManager.get('ceph-salt:bootstrap_mon_ip')
         if not bootstrap_mon_ip:
             return "No bootstrap Mon IP specified in config"
         if bootstrap_mon_ip in ['127.0.0.1', '::1']:
             return 'Mon IP cannot be the loopback interface IP'
+
+    # system_update
+    if not _is_boolean(PillarManager.get('ceph-salt:updates:enabled')):
```
bk201:
```suggestion
    if not isinstance(PillarManager.get('ceph-salt:updates:enabled'), bool)):
```

Also works.

bk201 (approved):

ceph_salt/validate/config.py
ricardoasmarques: Fixed

PR #220, opened 2020-05-07
Unable to see dashboard password

### LINKED ISSUE #192: Unable to see dashboard password

ricardoasmarques:
~~AFTER https://github.com/ceph/ceph-salt/pull/219~~

---

With this PR:

1) Default dashboard password will be generated by `ceph-salt`, so the user can see it in `config ls`
2) User will be forced to change Dashboard password on the first login
3) Default dashboard password is no longer hidden (`***`) on `config ls` because of 2)

Fixes: https://github.com/ceph/ceph-salt/issues/192

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -29,6 +29,7 @@ wait for other minions:
 {{ macros.begin_step('Run "cephadm bootstrap"') }}
 
 {% set dashboard_username = pillar['ceph-salt']['dashboard']['username'] %}
+{% set dashboard_password = pillar['ceph-salt']['dashboard']['password'] %}
 
 run cephadm bootstrap:
   cmd.run:
@@ -37,6 +38,10 @@ run cephadm bootstrap:
         cephadm --verbose bootstrap --mon-ip {{ pillar['ceph-salt']['bootstrap_mon_ip'] }} \
                 --config /tmp/bootstrap-ceph.conf \
                 --initial-dashboard-user {{ dashboard_username }} \
+                --initial-dashboard-password {{ dashboard_password }} \
+{%- if not pillar['ceph-salt']['dashboard']['password_update_required'] %}
+                --dashboard-password-noupdate \
+{%- endif %}
                 --output-keyring /etc/ceph/ceph.client.admin.keyring \
                 --output-config /etc/ceph/ceph.conf \
                 --skip-monitoring-stack \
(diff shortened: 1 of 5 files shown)


smithfarm:
so, if I **don't** explicitly set dashboard username/password via `ceph-salt config`, then ceph-salt generates a random password for me?

ricardoasmarques:
> so, if I **don't** explicitly set dashboard username/password via `ceph-salt config`, then ceph-salt generates a random password for me?

yes, when you use `ceph-salt config` for the first time, a randomly generated password is set by default

ricardoasmarques:
> 3. Default dashboard password is no longer hidden (***) on config ls because of 2)

I wonder if we should continue hiding the password on `ceph-salt config ls`, and have a `ceph-salt config /cephadm_bootstrap/dashboard/password info` command so see it in clear text.

ricardoasmarques:
> > so, if I **don't** explicitly set dashboard username/password via `ceph-salt config`, then ceph-salt generates a random password for me?
> 
> yes, when you use `ceph-salt config` for the first time, a randomly generated password is set by default

Or is it a better idea to always use `admin` as default password?

smithfarm:
My only concern is that, after all the changes are made and the dust has settled, we retain the ability to set "admin/admin" and have it stick and avoid having to change the password at first login. Ideally, if the user explicitly does:

```
ceph-salt config /cephadm_bootstrap/dashboard/username set admin
ceph-salt config /cephadm_bootstrap/dashboard/password set admin
```

in the config phase, then credentials will be "admin/admin" even after first login. But if the consensus is that users should jump through an additional hoop to avoid the forced password change at first login, that would be OK too.

ricardoasmarques:
@LenzGr @smithfarm 

Per our discussion today, the dashboard password behavior should be:

```
1) By default, `ceph-salt` generates a random dashboard password

2) By default, the user will be forced to change dashboard password on first login, but this can be changed on a new `/cephadm_bootstrap/dashboard/password_update_required` flag on `ceph-salt`

3) Password can be shown in clear text on `ceph-salt config ls` because by default, this passwords has to be changed on first login
```

Will update this PR as soon as possible.

smithfarm:
> 3) Password can be shown in clear text on `ceph-salt config ls` because by default, this passwords has to be changes on first login

Password can be shown in clear text **on screen and in log files**

ricardoasmarques:
PR is now updated according to https://github.com/ceph/ceph-salt/pull/220#issuecomment-628510241 :

```
master:~ # ceph-salt config /cephadm_bootstrap/dashboard ls
o- dashboard ...................................................... [...]
  o- force_password_update .................................... [enabled]
  o- password ................................................... [admin]
  o- username ................................................... [admin]
```

smithfarm (approved):
To preserve the current behavior, we'll need to add a commit to sesdev as well: https://github.com/SUSE/sesdev/pull/315

PR #224, opened 2020-05-11
Improve "ceph-salt config" performance

### LINKED ISSUE #222: "ceph-salt config" is slow on large clusters
With 104 nodes in the config, `ceph-salt` takes almost 3 minutes to open the shell:

```
master:~ # time ceph-salt config
/> exit

real	2m38.294s
user	0m28.353s
sys	0m1.543s
```

ricardoasmarques:
To improve `ceph-salt config` performance this PR will:

1) **Lazy load node grains**
       Only access grains that are needed, when needed.

2) **Read roles from pillar, not from grain**
       Pillar data is cached, and contains roles for all nodes, so we don't have to read grain on each node in the config.

Performance improvement on a `104 nodes` cluster:

**BEFORE**
```
master:~ # time ceph-salt config ls > /dev/null

real	2m33.535s
user	0m29.417s
sys	0m1.855s
```

**AFTER**
```
master:~ # time ceph-salt config ls > /dev/null

real	0m6.927s
user	0m5.687s
sys	0m0.362s
```

Fixes: https://github.com/ceph/ceph-salt/issues/222

### DIFF
--- a/ceph_salt/core.py
+++ b/ceph_salt/core.py
@@ -19,50 +19,74 @@ CEPH_SALT_GRAIN_KEY = 'ceph-salt'
 class CephNode:
     def __init__(self, minion_id):
         self.minion_id = minion_id
-        self.hostname = None
-        self.roles = None
-        self.execution = {}
-        self.public_ip = None
-        self.subnets = None
-        self.public_subnet = None
-        self._load()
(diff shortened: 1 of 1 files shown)


smithfarm (approved):
LGTM

PR #228, opened 2020-05-12
salt_utils: don't log dashboard password

tserong:


ricardoasmarques (approved):
lgtm

ricardoasmarques:
A similar change is required for the `set` method, but it's now included in https://github.com/ceph/ceph-salt/pull/234

PR #233, opened 2020-05-14
chrony_sync_clock: wait longer for sync

smithfarm:
I had a test fail because it exhausted all 30 tries without successfully
syncing with the remote time server.


ricardoasmarques (approved):
lgtm

PR #234, opened 2020-05-14
Do not log pillar data secrets

### LINKED ISSUE #229: `ceph-salt export` logs private ssh key and dashboard password to /var/log/ceph-salt.log
When running `ceph-salt config ls`, the private ssh key is (correctly) not included in the log by the PillarManager.get() function, which looks like it's being called for every value individually (the dashboard password will also be skipped once #228 goes in). However, when running `ceph-salt export`, it calls PillarManager.get('ceph-salt'), which dumps the entire pillar contents, so everything in the pillar, including the private ssh key and dashboard password land in /var/log/ceph-salt.log

ricardoasmarques:
Passwords and private keys must be omitted/hidden in the log, even when pillar data result is a `dict` (e.g. when getting `ceph-salt`, `ceph-salt:ssh` or `ceph-salt:dashboard`).

Note that this applies to pillar data **reads** (e.g. `ceph-salt export`) and **writes** (e.g. `ceph-salt import`).

Fixes: https://github.com/ceph/ceph-salt/issues/229

### DIFF
--- a/ceph_salt/salt_utils.py
+++ b/ceph_salt/salt_utils.py
@@ -1,4 +1,5 @@
 import contextlib
+import copy
 import logging
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


tserong (approved):
LGTM!

PR #235, opened 2020-05-15
Add 'cephadm' role

### LINKED ISSUE #217: Need to be able to have the master node *not* managed by cephadm
I believe that in most cases it's inadvisable to have the master node also run ceph daemons, because that would make one of your cluster nodes also the salt-master, which suddenly makes that node "special" and different somehow. This is probably especially true where the salt-master node is also running SUMA or other management tooling that's well outside Ceph's domain.

Right now however, all nodes are added to cephadm's control, so when you deploy some more mons or mgrs after initial bootstrap, some may end up running on the master node. I think we need a way to disable this, i.e. keep the master in the list of minions (so it can be set as the time server), but skip the `ceph orch host add` step for the master node.

ricardoasmarques:
This PR adds a new `cephadm` role.

A minion with `cephadm` role is a node that will be controlled by cephadm/ceph orchestrator.

Fixes: https://github.com/ceph/ceph-salt/issues/217

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,15 +50,23 @@ set the initial deployment of your cluster:
 ceph-salt config
 ```
 
-First step of configuration is to add the salt-minions that should be used for
-deploying Ceph the command `add` under `/ceph_cluster/minions` option supports
-autocomplete and glob expressions:
+First step of configuration is to add the salt-minions that should be managed
+by`ceph-salt`.
+The `add` command under `/ceph_cluster/minions` option supports autocomplete
+and glob expressions:
 
 ```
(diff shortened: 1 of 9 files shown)


tserong (commented):
Looks pretty good, but probably also wants an addition to README.md to mention the new role

ricardoasmarques:
> Looks pretty good, but probably also wants an addition to README.md to mention the new role

Good catch. I've now updated the `README` to include the new role.

tserong (approved):

smithfarm (approved):

PR #244, opened 2020-05-20
Persist journal logs

### LINKED ISSUE #240: Create `/var/log/journal` directory
In order to have persistent logs, `ceph-salt` should create `/var/log/journal` directory in all minions, during the provisioning.

ricardoasmarques:
Create `/var/log/journal` directory, so journal logs will be persisted.

Fixes: https://github.com/ceph/ceph-salt/issues/240

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/software.sls
+++ b/ceph-salt-formula/salt/ceph-salt/software.sls
@@ -10,6 +10,14 @@ install required packages:
       - podman
     - failhard: True
 
+/var/log/journal:
+  file.directory:
+    - user: root
+    - group: root
+    - mode: '0755'
+    - makedirs: True
+    - failhard: True
+
 {% if pillar['ceph-salt']['updates']['enabled'] %}
 
 {{ macros.begin_step('Update all packages') }}


smithfarm (approved):

smithfarm (changes requested):
I looked up how this is supposed to be done by the end user. The instructions are here:

https://doc.opensuse.org/documentation/leap/reference/html/book.opensuse.reference/cha-journalctl.html#sec-journalctl-persistent

When I tried this on my machine, systemd created `/var/log/journal` with mode 0755:

```
# ls -ld /var/log/journal
drwxr-xr-x 3 root root 46 May 20 13:12 /var/log/journal
```

So I'd suggest to use 0755 instead of 0770, and it would probably also be a good idea to sed `/etc/systemd/journald.conf` so it has `Storage=persistent`.

tserong:
I agree with the permissions change @smithfarm suggested after doing a couple of quick tests here.

I'm not sure about sedding /etc/systemd/journald.conf though.  By default, everything seems to be commented out, which means Storage=auto applies, so it'll be persistent provided /var/log/journal exists. If somehow Storage=[anything other than auto], that presumably means the admin has specifically decided they want non-default behaviour, so maybe we shouldn't screw around with it?

ricardoasmarques:
@smithfarm @tserong I've now changed permissions to `0755`.

Regarding `/etc/systemd/journald.conf`, I agree that we can assume the default. And the detault `Storage=auto` works fine for us https://www.freedesktop.org/software/systemd/man/journald.conf.html

tserong (approved):

smithfarm (approved):

Issue #241, opened 2020-05-19
Excessive timeout on failed time sync


swiftgist:
Ten minutes is too long for a failure

```
# time /tmp/chrony_sync_clock.sh 
200 OK
200 OK
try: 1, refid: 00000000, correction: 0.000000000, skew: 0.000
try: 2, refid: 00000000, correction: 0.000000000, skew: 0.000
try: 3, refid: 00000000, correction: 0.000000000, skew: 0.000
...
try: 59, refid: 00000000, correction: 0.000000000, skew: 0.000
try: 60, refid: 00000000, correction: 0.000000000, skew: 0.000

real    10m5.624s
user    0m0.011s
sys     0m0.029s


```

I suggest setting the command to `chronyc waitsync 6 0.04` or removing the waitsync since the previous `chronyc makestep` can fail already.

smithfarm:
May I ask why you think 10 minutes too long? The waitsync has been known to take several minutes to complete successfully.

The purpose of syncing the clocks is two-fold:

1. to ensure that the cluster does not come up in HEALTH_WARN due to clock skew
2. to flag time sync issues before the cluster gets deployed

To me, 10 minutes does not seem like a very long time for a command that is expected to only run once? The argument for waiting up to 10 minutes is that we need to be fairly certain that there really is a problem with the external time server specified by the user before failing `ceph-salt apply` because we _think_ it doesn't work.

smithfarm:
Maybe we can do some sanity checks on the external time server(s) specified by the user, before triggering the clock sync script? Then if someone specifies e.g. "pool.npt.org" (i.e. a typo), we could fail much quicker.

swiftgist:
> May I ask why you think 10 minutes too long? The waitsync has been known to take several minutes to complete successfully.

For the general case, ten minutes is a long time for failure when most/many users may simply rerun `ceph-salt` without investigation to see if the problem is temporary.  With further investigation and running the script directly, the additional runs do add up.

Is your experience with the "known to take several minutes" only in the testing environment?   I would be surprised in general that `chronyc makestep` is insufficient.

swiftgist:
> Maybe we can do some sanity checks on the external time server(s) specified by the user, before triggering the clock sync script? Then if someone specifies e.g. "pool.npt.org" (i.e. a typo), we could fail much quicker.

[...]

PR #245, opened 2020-05-20
Sanity check external time servers before applying config

smithfarm:
References: https://github.com/ceph/ceph-salt/issues/241

### DIFF
--- a/ceph_salt/validate/config.py
+++ b/ceph_salt/validate/config.py
@@ -1,4 +1,34 @@
+import socket
+import subprocess
+
 from ..salt_utils import PillarManager
+from ..exceptions import CmdException
+
+
+def host_resolvable_to_ip(host):
+    return list(set(map(
+        lambda a: a[4][0],
+        socket.getaddrinfo(host, 123)
+    )))
+
+
+def run_sync(command, cwd=None):
+    # logger.info("Running sync command (%s): %s", cwd if cwd else ".", command)
+    with subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=cwd) as proc:
+        stdout, stderr = proc.communicate()
+        if proc.returncode != 0:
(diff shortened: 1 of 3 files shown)


smithfarm:
@swiftgist This draft PR adds a sanity check of the external time server hostnames provided by the user, which should make us fail fast in the case when the user specifies a hostname that isn't actually a functioning time server.

ceph_salt/validate/config.py
```diff
@@ -66,6 +96,19 @@ def validate_config(host_ls):
             return 'No external time servers specified in config'
         if not time_server_is_minion and external_time_servers:
             return not_minion_err.format('external time servers')
+        # assert that external_time_servers are sane
+        if external_time_servers:
+            external_time_server_ip_addresses = []
+            ahost = None
+            for ahost in external_time_servers:
+                ipaddrs = host_resolvable_to_ip(ahost)
```
ricardoasmarques, line 104: I'm afraid this is more complex than anticipated.

This PR is checking if `salt-master` (the host where `ceph-salt` is running) can resolve `external time servers`, but the host that should be able to resolve it is the `time_server_minion`, not the master.

I wonder if salt `network` module can help here https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.network.html

Some examples:

```
salt '*' network.ping pool.ntp.org return_boolean=True

salt '*' network.connect pool.ntp.org port=123 proto=udp
```

Example of how to execute a salt module on a minion (from `ceph-salt` python code):
```
returns = SaltClient.local().cmd(time_server_host, 'network.connect', [ahost, 'port=123', 'proto=udp'])
```

But for some reason, `salt '*' network.connect pool.ntp.org port=123 proto=udp` is always failing on my tests, so I wonder if the best we can to is simply check if `ping` works? Or maybe we can't assume that external time servers will reply to `ping` request?

smithfarm:
will take a completely different approach

Issue #236, opened 2020-05-17
Need to invoke `ceph orch apply crash`


tserong:
We need to add a call to `ceph orch apply crash` to make the orchestrator deploy the ceph-crash service. Without this, ceph crashes never get posted, so won't appear in `ceph crash stat` or `ceph crash ls` (`cephadm bootstrap` would normally do this, but we're calling it with `--skip-ssh`, so don't get this behaviour by default - see also https://tracker.ceph.com/issues/45572).

sebastian-philipp:
I personally don't like to simply run `ceph orch apply crash` without any placement specs. 

I know how it works, but it I doubt that everyone knows the default placement specs for each service.

I'd like to make `--placement` mandatory and force users to run `ceph orch apply crash --placement '*'`

tserong:
I agree that nobody will know WTF the default placement specs are, and users should explicitly specify a placement when they deploy a service. For the crash service though, I still think we should automate this, i.e. ceph-salt should run `ceph orch apply crash '*'`, without the user having to do it by hand.

sebastian-philipp:
which is already the caes, as crash is automatically enabled if you don't say --skip-ssh

Maybe we should change the behavior of --skip-ssh to align it with expectations?

I.e. why skipping `apply crash '*'` if we don't have any hosts, that's unrelated

tserong:
Yeah, but ceph-salt *does* say --skip-ssh :-)

--skip-ssh also skips `ceph mgr module enable cephadm`, and `ceph orch set backend cephadm`. So maybe --skip-ssh should be changed to *only* skip ssh, and still have `cephadm boostrap` enable the cephadm module? In which case, ceph-salt wouldn't need to do those latter things anymore.

sebastian-philipp:
yes yes yes yes

ricardoasmarques:
@tserong @sebastian-philipp FYI, I've submitted a `cephadm` issue - https://tracker.ceph.com/issues/45629

Issue #173, opened 2020-04-09
Should be possible to re-apply config on all minions


ricardoasmarques:
ATM, after initial deployment, configuration can only be applied to one minion at a time:

 ```
ceph-salt deploy <minion_id>
 ```

Depending on the configuration we change, this may result in an inconsistent deployment (e.g SSH keys changed, or time server changed).

Should be possible to re-apply the whole config after initial deployment, by running:
 ```
ceph-salt deploy
 ```

Note that `Reboot` must be skipped for nodes that are already part of the Ceph cluster[1][2]. The following message should be shown in this case:
```
"Minion is already managed by cephadm, please reboot manually"
```

[1] maybe we can use `podman ps` to check if any ceph container is running?
[2] also, we should check if any ceph service is running (in case we are dealing with an upgrade from nautilus to octopus)?

---

Relates to https://tracker.ceph.com/issues/44825
Relates to https://github.com/ceph/ceph-salt/issues/176

sebastian-philipp:
FYI, there is a demand from Ansible to make `bootstrap` idempotent: https://tracker.ceph.com/issues/44825

PR #250, opened 2020-05-26
Rely on "bootstrap" to configure MGR module (and more...)

ricardoasmarques:
~~**After https://github.com/ceph/ceph/pull/35195**~~
~~**After https://github.com/ceph/ceph/pull/35347**~~

---

This PR contains two commits, each commit fixes a different issue.

**1. Need to invoke `ceph orch apply crash`** (https://github.com/ceph/ceph-salt/issues/236)

We will now rely on `cephadm bootstrap` to configure `cephadm` MGR module, which will guarantee that all initializations are properly done (including the execution of `ceph orch apply crash`)

**2. Make bootstrap_mon optional** (https://github.com/ceph/ceph-salt/issues/265)

It's now possible to explicitly say that we don't want `ceph-salt` to run `cephadm bootstrap` by setting `ceph-salt config /cephadm_bootstrap disable`, so the new behavior is:
1) `bootstrap_minion` is required for initial `ceph-salt apply` execution but can be removed latter if needed/wanted.
2) If `bootstrap is enabled`, then `/cephadm_bootstrap/*` options are required (including `mon_ip`). In this scenario `cephadm bootstrap` will be executed on the `bootstrap_minion`.
3) If `bootstrap is disabled`, then `/cephadm_bootstrap/*` options will be ignored and `ceph-salt` will assume that `bootstrap_minion` has access to a running cluster (e.g. an upgrade from Nautilus to Octopus). In this scenario `ceph-salt` will simply configure/enable cephadm MGR module.

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -47,7 +47,8 @@ run cephadm bootstrap:
                 --skip-monitoring-stack \
                 --skip-prepare-host \
                 --skip-pull \
-                --skip-ssh \
+                --ssh-private-key /tmp/ceph-salt-ssh-id_rsa \
+                --ssh-public-key /tmp/ceph-salt-ssh-id_rsa.pub \
 {%- for arg, value in pillar['ceph-salt'].get('bootstrap_arguments', {}).items() %}
                 --{{ arg }} {{ value if value is not none else '' }} \
 {%- endfor %}
@@ -60,21 +61,6 @@ run cephadm bootstrap:
     - failhard: True
 
 {{ macros.end_step('Run "cephadm bootstrap"') }}
-
-{{ macros.begin_step('Configure cephadm MGR module') }}
-
-configure ssh orchestrator:
-  cmd.run:
-    - name: |
-        ceph config-key set mgr/cephadm/ssh_identity_key -i /tmp/ceph-salt-ssh-id_rsa
-        ceph config-key set mgr/cephadm/ssh_identity_pub -i /tmp/ceph-salt-ssh-id_rsa.pub
-        ceph mgr module enable cephadm && \
-        ceph orch set backend cephadm
-    - onchanges:
-      - cmd: run cephadm bootstrap
-    - failhard: True
-
-{{ macros.end_step('Configure cephadm MGR module') }}
 
 {{ macros.end_stage('Bootstrap the Ceph cluster') }}
 
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -1,6 +1,6 @@
 {% import 'macros.yml' as macros %}
 
-{% if grains['id'] == pillar['ceph-salt']['bootstrap_minion'] %}
+{% if grains['id'] == pillar['ceph-salt'].get('bootstrap_minion') %}
 
 {{ macros.begin_stage('Bootstrap the Ceph cluster') }}
 
@@ -25,6 +25,8 @@ wait for other minions:
     - hosts: {{ pillar['ceph-salt']['minions']['all'] }}
     - failhard: True
 {{ macros.end_step('Wait for other minions') }}
+
+{% if pillar['ceph-salt']['bootstrap_enabled'] %}
 
 {{ macros.begin_step('Run "cephadm bootstrap"') }}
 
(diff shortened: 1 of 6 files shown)


ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
```diff
@@ -61,21 +62,6 @@ run cephadm bootstrap:
 
 {{ macros.end_step('Run "cephadm bootstrap"') }}
 
-{{ macros.begin_step('Configure cephadm MGR module') }}
```
ricardoasmarques, line 64: We may need to keep `configure ssh orchestrator` due to https://github.com/ceph/ceph-salt/pull/267.

Perhaps, run it on admin nodes[1] to guarantee that we don't require a `bootstrap_minion` when upgrading from `nautilus`. 
 
[1] https://github.com/ceph/ceph-salt/blob/master/ceph-salt-formula/salt/ceph-salt/ceph-admin.sls

tserong (commented):
I'm OK with this in general, but, yeah, we still need the orchestrator enablement for the upgrade case (see https://github.com/ceph/ceph-salt/pull/267#issuecomment-644558290)

ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
```diff
@@ -70,12 +75,13 @@ configure ssh orchestrator:
         ceph config-key set mgr/cephadm/ssh_identity_pub -i /tmp/ceph-salt-ssh-id_rsa.pub
         ceph mgr module enable cephadm && \
         ceph orch set backend cephadm
```
ricardoasmarques, line 77: @tserong do we also need to tun `ceph orch apply crash` here?

ricardoasmarques:
I've now pushed a new commit and updated PR description accordingly.

@tserong please check if this PR fixes  https://github.com/ceph/ceph-salt/issues/236 and https://github.com/ceph/ceph-salt/pull/267

@swiftgist please check if this PR is sufficient to fix https://github.com/ceph/ceph-salt/issues/265

ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
swiftgist: The configure cephadm mgr module has also been a part of the fresh deployment.  Won't this still be necessary?

swiftgist:
Without a boostrap mon, should `ceph-salt` run the related module and backend commands?  For an upgrade scenario, this would mean that the mon/mgr need to be upgraded _prior_ to running `ceph-salt`.  Considering that `ceph-salt` is primarily for ensuring prerequisites, I do not think this is obvious.

ricardoasmarques:
> Without a boostrap mon, should `ceph-salt` run the related module and backend commands? For an upgrade scenario, this would mean that the mon/mgr need to be upgraded _prior_ to running `ceph-salt`. Considering that `ceph-salt` is primarily for ensuring prerequisites, I do not think this is obvious.

My understanding from https://github.com/ceph/ceph-salt/pull/267 was that, for an upgrade scenario, we should do everything except the `cephadm bootstrap` part. Let's wait for @tserong take on this so we can adapt `ceph-salt` to behave in a way that makes upgrades easier.

ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
ricardoasmarques: The current implementation assumes that:

1) The first time we execute `ceph-salt apply`, the `bootstrap_minion` is required. This applies to fresh install and upgrades from Nautilus.

2) After first `ceph-salt apply`execution, there's no need to keep `bootstrap_minion` because it's no longer needed (but we can keep it if we want).

So in summary, this role is only needed for the first execution.

ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
tserong: > @tserong do we also need to tun `ceph orch apply crash` here?

I don't think so. We *do* ultimately need `ceph orch apply crash` to happen in the upgrade case, but I don't think this is the right place for it. We actually need to apply service specs for all adopted services, and can probably apply the crash service when doing that, but this is outside the scope of what ceph-salt does so far (if that changes, we can always add it in later)

tserong:
> @tserong please check if this PR fixes #236 and #267

From reviewing the code, it looks like this'll do the right thing (I'd still like to try running it to test).

> Without a boostrap mon, should `ceph-salt` run the related module and backend commands? For an upgrade scenario, this would mean that the mon/mgr need to be upgraded _prior_ to running `ceph-salt`. Considering that `ceph-salt` is primarily for ensuring prerequisites, I do not think this is obvious.

I agree, it's not obvious - the upgrade scenario is irritatingly chicken-and-eggy. Maybe there's a better approach to clear this up somehow, but IMO it's OK to take this PR as-is, and consider further upgrade-related reworking as a separate issue.

ricardoasmarques:
If we want to support re-applying config as describes in https://github.com/ceph/ceph-salt/issues/173 then we should not rely on `cephadm bootstrap` to configure keys because if the user changes those keys we need them to be updated next time `ceph-salt apply` is executed.

So I'm closing this PR for now and will open a new one.

tserong:
Excellent point... Although maybe it makes sense to still rely on `cephadm bootstrap` to initially configure keys (because it's easy, and also because bootstrap sets up the crash service), and we only need to add an extra step to re-apply keys to a running system?

ricardoasmarques:
> Excellent point... Although maybe it makes sense to still rely on `cephadm bootstrap` to initially configure keys (because it's easy, and also because bootstrap sets up the crash service), and we only need to add an extra step to re-apply keys to a running system?

Yes, I'll consider that in my new PR, thanks.

PR #254, opened 2020-05-28
Switch to reactive UI refresh after execution is complete

### LINKED ISSUE #226: Add ability to pause interactive screen refreshes
When running `ceph apply` interactively, it may be helpful to pause screen refreshes. This would aid in analyzing error messages, particularly in large environments. Ideally, this would also allow copy and paste from the interactive shell output.

If this is possible, it should be added as "(p)ause" (similar to the already existing "(c)ollapse" option). (Although there is a risk that people will assume that implies suspending all operations.)

ricardoasmarques:
After execution is complete we no longer display the loading icon, so we can switch from "periodic refresh" to "reactive refresh" (only refresh UI when needed / user action is performed).

This will allow the user to `Select` and `Copy` text from the UI after execution completes (which is not possible if UI keeps refreshing periodically):

Screenshot from 2020-05-28 14-31-57

Fixes: https://github.com/ceph/ceph-salt/issues/226

### DIFF
--- a/ceph_salt/apply.py
+++ b/ceph_salt/apply.py
@@ -243,7 +243,7 @@ class CursesScreen:
         try:
             ch = self.stdscr.getch()
             if ch == -1:
-                return
+                return False
             if ch == curses.KEY_NPAGE:
                 if self.body:
                     if self.body_pos < self.body_current_row - self.body_height:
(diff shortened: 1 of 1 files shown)


smithfarm:
I can confirm that this PR stops the screen from refreshing as soon as execution stops, but I can't see how it implements the "Pause" functionality that #226 requested?

ricardoasmarques:
> I can confirm that this PR stops the screen from refreshing as soon as execution stops, but I can't see how it implements the "Pause" functionality that #226 requested?

@smithfarm I've now pushed a new commit with the new `"(p) pause"` command.

smithfarm (approved):
LGTM

PR #257, opened 2020-06-01
Install ceph-salt ssh keys on admin minions

### LINKED ISSUE #191: Install ssh keys in /root/.ssh
In ceph-salt-15.2.1+1587037332.ge798398-1.1.noarch, the ssh keys are placed in /tmp as /tmp/ceph-salt-ssh-id_rsa and /tmp/ceph-salt-ssh-id_rsa.pub.

In a fresh deployment scenario, trying to ssh from an admin node to another prompts for a password as configured.  Considering the /root/.ssh/authorized_keys is updated correctly, I suggest installing the keys in /root/.ssh with the default names.

With regards to upgrades or possibly existing keys, only populate the files if the pillar contents exist.

ricardoasmarques:
With this PR, SSH keys will be installed in `/root/.ssh` (instead of `/tmp`), and if the node has the `admin` role, then those keys will not be removed after execution.

The user will then be able to SSH from one `admin` node to any other node by running:

```
master:~ # ssh node2 -i ~/.ssh/ceph-salt
Last login: Mon Jun  1 15:35:54 2020 from 10.20.162.200
Have a lot of fun...
node2:~ # 
```

~~Note that we are not using the default name (`id_rsa`) to guarantee that we will not override existing keys.~~

Fixes: https://github.com/ceph/ceph-salt/issues/191

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -36,8 +36,9 @@ def get_remote_grain(host, grain):
     """
     ssh_user = __pillar__['ceph-salt']['ssh']['user']
     sudo = 'sudo ' if ssh_user != 'root' else ''
+    home = '/home/{}'.format(ssh_user) if ssh_user != 'root' else '/root'    
     ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
-                                  "-i /tmp/ceph-salt-ssh-id_rsa {}@{} "
+                                  "-i {}/.ssh/ceph-salt {}@{} "
                                   "\"{}python3 - <<EOF\n"
                                   "import json\n"
                                   "import salt.utils.data\n"
@@ -46,7 +47,7 @@ def get_remote_grain(host, grain):
                                   "    grains = yaml.full_load(grains_file)\n"
                                   "val = salt.utils.data.traverse_dict_and_list(grains, '{}')\n"
                                   "print(json.dumps({{'local': val}}))\n"
-                                  "EOF\"".format(ssh_user, host, sudo, grain))
+                                  "EOF\"".format(home, ssh_user, host, sudo, grain))
     if ret['retcode'] != 0:
         return None
     return json.loads(ret['stdout'])['local']
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -38,7 +38,7 @@ def get_remote_grain(host, grain):
     sudo = 'sudo ' if ssh_user != 'root' else ''
     home = '/home/{}'.format(ssh_user) if ssh_user != 'root' else '/root'    
     ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
-                                  "-i {}/.ssh/ceph-salt {}@{} "
+                                  "-i {}/.ssh/id_rsa {}@{} "
                                   "\"{}python3 - <<EOF\n"
                                   "import json\n"
                                   "import salt.utils.data\n"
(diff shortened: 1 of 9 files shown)


ricardoasmarques:
Since we are no longer using `root` user by default, this PR is now using default key file names (`~/.ssh/id_rsa.pub` and `~/.ssh/id_rsa.pub`)

ceph-salt-formula/salt/_modules/ceph_salt.py
```diff
@@ -36,8 +36,9 @@ def get_remote_grain(host, grain):
     """
     ssh_user = __pillar__['ceph-salt']['ssh']['user']
     sudo = 'sudo ' if ssh_user != 'root' else ''
+    home = '/home/{}'.format(ssh_user) if ssh_user != 'root' else '/root'    
     ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
-                                  "-i /tmp/ceph-salt-ssh-id_rsa {}@{} "
+                                  "-i {}/.ssh/id_rsa {}@{} "
```
smithfarm, line 41: can we just drop `-i {}/.ssh/id_rsa` altogether, since it is the default?

ceph-salt-formula/salt/_states/ceph_orch.py
```diff
@@ -29,12 +29,14 @@ def set_admin_host(name, timeout=1800):
             if provisioned:
                 ssh_user = __pillar__['ceph-salt']['ssh']['user']
                 sudo = 'sudo ' if ssh_user != 'root' else ''
+                home = '/home/{}'.format(ssh_user) if ssh_user != 'root' else '/root'                
                 status_ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
-                                                     "-i /tmp/ceph-salt-ssh-id_rsa {}@{} "
+                                                     "-i {}/.ssh/id_rsa {}@{} "
```
smithfarm: can we just drop `-i {}/.ssh/id_rsa` altogether, since it is the default?

ceph-salt-formula/salt/_states/ceph_orch.py
```diff
@@ -63,12 +65,14 @@ def wait_until_ceph_orch_available(name, timeout=1800):
         admin_host = __salt__['grains.get']('ceph-salt:execution:admin_host')
         ssh_user = __pillar__['ceph-salt']['ssh']['user']
         sudo = 'sudo ' if ssh_user != 'root' else ''
+        home = '/home/{}'.format(ssh_user) if ssh_user != 'root' else '/root'    
         status_ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
-                                             "-i /tmp/ceph-salt-ssh-id_rsa {}@{} "
+                                             "-i {}/.ssh/id_rsa {}@{} "
```
smithfarm, line 78: can we just drop `-i {}/.ssh/id_rsa` altogether, since it is the default?

smithfarm:
@ricardoasmarques I noticed that the following code is duplicated in a bunch of places:

```
"ssh -o StrictHostKeyChecking=no "
"-i {}/.ssh/id_rsa {}@{} "
```

Could this string be templated so we have the code in only one place within the Salt Formula?

ricardoasmarques:
> "ssh -o StrictHostKeyChecking=no "

@smithfarm  I agree, but I've created a separate issue so we can implement it on a dedicated PR https://github.com/ceph/ceph-salt/issues/360

ceph-salt-formula/salt/_states/ceph_orch.py
ricardoasmarques: I think we cant, see:

```
master:~ # ssh -o StrictHostKeyChecking=no ceph-salt@node1.pacific.test pwd
Password: 

master:~ # ssh -o StrictHostKeyChecking=no -i /home/ceph-salt/.ssh/id_rsa ceph-salt@node1.pacific.test pwd
/home/ceph-salt
```

ceph-salt-formula/salt/_modules/ceph_salt.py
smithfarm: no, because Salt runs the command as root

ceph-salt-formula/salt/_states/ceph_orch.py
smithfarm: no, because Salt runs the command as root

smithfarm (approved):

PR #259, opened 2020-06-02
Add "Conflicts: deepsea" to spec file

tserong:
This ensures that when installing ceph-salt, deepsea will be removed (`zypper in` will say something like "Problem: ceph-salt conflicts with deepsea - solution 1: deinstallation of deepsea, solution 2: do not install ceph-salt").


tserong:
jenkins failure looks unrelated to me

smithfarm (approved):

PR #262, opened 2020-06-02
Allow users to disable custom registries configuration

ricardoasmarques:
This PR is an alternative to https://github.com/ceph/ceph-salt/pull/214

---

With this PR, users will be able to control if `/etc/containers/registries.conf` should be managed by `ceph-salt` or not.

By default, `/etc/containers/registries.conf` management is enabled, but can be disabled by:
```
# ceph-salt config /containers/registries_conf disable
```

(Note that we had to move `/containers/registries` to `/containers/registries_conf/registries`)

### DIFF
--- a/ceph_salt/config_shell.py
+++ b/ceph_salt/config_shell.py
@@ -388,31 +388,38 @@ CEPH_SALT_OPTIONS = {
                     },
                 }
             },
-            'registries': {
-                'type': 'list_dict',
-                'help': '''
-                        List of custom registries in v2 format.
-                        =======================================
+            'registries_conf': {
+                'type': 'group',
+                'help': "Custom registries configuration",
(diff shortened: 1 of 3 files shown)


ceph_salt/config_shell.py
```diff
@@ -388,31 +388,38 @@ def generate_password():
...
-                        'required': True
-                    },
-                    'prefix': {},
-                    'insecure': {
-                        'validator': BooleanStringValidator,
-                        'transformer': BooleanStringTransformer
+            'registries_conf': {
+                'type': 'group',
+                'help': "Custom registries configuration",
+                'handler': FlagGroupHandler('ceph-salt:container:registries_enabled', True),
```
bk201, line 394: Maybe disabling it by default?

bk201 (approved):

toabctl (approved):

PR #126, opened 2020-03-12
configshell: Handle execution errors (e.g. Command not found)

ricardoasmarques:
Handle `configshell` execution errors.

Screenshot from 2020-03-12 14-58-51

Fixes: https://github.com/ceph/ceph-salt/issues/66

Signed-off-by: Ricardo Marques <rimarques@suse.com>

PR #264, opened 2020-06-04
configshell: Handle execution errors on command line

tserong:
Similar to https://github.com/ceph/ceph-salt/pull/126, but for running individual commands, so instead of a backtrace we get a nice error e.g.:

```
master:~ # ceph-salt config foo
Command not found foo
```


ricardoasmarques (approved):
lgtm

Issue #265, opened 2020-06-05
Make bootstrap_mon optional


swiftgist:
For considering upgrades where a Ceph cluster already exists, I think it makes sense to use `ceph-salt apply` for the subset of states that are not related to a fresh deployment.  These are 

```
    - .sshkey
    - .software
    - .container
    - .apparmor
    - .time-prep
    - .time-sync
    - .cephtools
```

If the remaining steps are dependent on the presense of a bootstrap_mon, then

```
    - .cephbootstrap
    - .find-admin-host
    - .cephorch
    - .ceph-admin
```
would run for a fresh deployment.  All other scenarios would skip these parts.

Note that `cephtools` above does have a bootstrap_mon condition for setting the permissions on `/var/log/ceph` that would need to be addressed.

tserong:
We still need the "configure ssh orchestrator" part of cephbootstrap for the upgrade case (see #267)

ricardoasmarques:
Checking for `bootstrap_minion` presence to control if those steps are executed or not will force us to always have a `bootstrap_minion` in our config, even after the initial deployment, otherwise new minions will not be automatically added to the orchestrator.

If we want to keep the behavior of allowing the user to remove the bootstrap minion after day 1, then we need a new setting on `ceph-salt config` to control if those steps should be executed or not.

Not sure what is the best approach here.

PR #267, opened 2020-06-09
Use `ceph orch status` as check for configuring ssh orchestrator

tserong:
Previously, orchestrator configuration would only happen if `cephadm bootstrap` was actually executed (the "onchanges" condition).  This creates a problem for upgrades from DeepSea to ceph-salt, where `cephadm bootstrap` will never run (because /etc/ceph/ceph.conf and /etc/ceph/ceph.client.admin.keyring already exist), which in turn means the orchestrator configuration never happens.

By replacing "onchanges" with "unless: ceph orch status...", we decouple these two steps, so the orchestrator configuration will happen in any case where the backend is not already set to cephadm.

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -70,8 +70,7 @@ configure ssh orchestrator:
         ceph config-key set mgr/cephadm/ssh_identity_pub -i /tmp/ceph-salt-ssh-id_rsa.pub
         ceph mgr module enable cephadm && \
         ceph orch set backend cephadm
-    - onchanges:
-      - cmd: run cephadm bootstrap
+    - unless: "ceph orch status 2>&1 | grep -q 'Backend: cephadm'"
     - failhard: True
 
 {{ macros.end_step('Configure cephadm MGR module') }}


smithfarm:
@tserong I'm curious, how is the bootstrap_minion grain set in the upgrade scenario? I ask, because almost all of ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls is enclosed in the following conditional:

```
{% if grains['id'] == pillar['ceph-salt']['bootstrap_minion'] %}
...
{% endif %}
```

I guess you could find out which minions have both a MON and a MGR running on them and randomly choose one of those as the "bootstrap minion". Still, it seems odd to have a "bootstrap minion" in an upgrade scenario where bootstrap is not being run?

An alternative might be to split the `configure ssh orchestrator` step into a separate SLS file?

tserong:
> @tserong I'm curious, how is the bootstrap_minion grain set in the upgrade scenario?

Right now, the admin has to set it. I've just been setting it to one of the mons in my testing, and because there's already a ceph.conf and admin keyring on that node, the actual `cephadm bootstrap` invocation doesn't happen. Somewhat counterintuitive, but it works.

> An alternative might be to split the `configure ssh orchestrator` step into a separate SLS file?

Yeah, that would probably make a bit more sense, and ties in with #265.

Edit: Oh, wait, except, the `configure ssh orchestrator` step really only wants to be run on one node, and it has to be a node with the ceph.conf and admin keyring. So maybe it does make sense to still nominate a bootstrap node?

ricardoasmarques:
@tserong @smithfarm Note that this PR will conflict with https://github.com/ceph/ceph-salt/pull/250 where `configure ssh orchestrator` is being removed.

In https://github.com/ceph/ceph-salt/pull/250 we are relying on cephadm to `configure ssh orchestrator` because since https://github.com/ceph/ceph/pull/35195 it's now possible to provide ssh keys during `cephadm bootstrap`.

ricardoasmarques:
> Edit: Oh, wait, except, the `configure ssh orchestrator` step really only wants to be run on one node, and it has to be a node with the ceph.conf and admin keyring. So maybe it does make sense to still nominate a bootstrap node?

If `configure ssh orchestrator` is idempotent, what about running it on admin nodes[1]?

[1] https://github.com/ceph/ceph-salt/pull/250/files#r440114213

tserong:
> If `configure ssh orchestrator` is idempotent, what about running it on admin nodes[1]?

That should work fine. With the ssh change from #250 it would probably only need to be:

```
configure ssh orchestrator:
  cmd.run:
    - name: ceph mgr module enable cephadm && ceph orch set backend cephadm
    - unless: "ceph orch status 2>&1 | grep -q 'Backend: cephadm'"
    - failhard: True
```

The `ceph orch status` check means it won't re-apply if it's already been done (either by boostrap, or on some other admin node, in case there are multiple admin nodes). Even if two admin nodes somehow both ran the above at exactly the same time, it should *still* be fine, because there's no harm in enabling the same module multiple times (you'll just get back "module 'cephadm' is already enabled")

tserong:
> If `configure ssh orchestrator` is idempotent, what about running it on admin nodes[1]?

That should work fine. The `ceph orch status` check means it won't re-apply if it's already been done (either by boostrap, or on some other admin node, in case there are multiple admin nodes). Even if two admin nodes somehow both ran it at exactly the same time, it should still be fine, because there's no harm in enabling the same module multiple times (you'll just get back "module 'cephadm' is already enabled")

I originally thought we cut it down to just running "ceph mgr module enable cephadm && ceph orch set backend cephadm", but of course we're still going to need the ssh key setup for the upgrade case as well.

ricardoasmarques:
> > If `configure ssh orchestrator` is idempotent, what about running it on admin nodes[1]?
> 
> That should work fine. The `ceph orch status` check means it won't re-apply if it's already been done (either by boostrap, or on some other admin node, in case there are multiple admin nodes). Even if two admin nodes somehow both ran it at exactly the same time, it should still be fine, because there's no harm in enabling the same module multiple times (you'll just get back "module 'cephadm' is already enabled")

As an alternative, we can run it only on the "first" admin only, something like `if minion_id == pillar['ceph-salt']['minions']['admin][0]` (not sure if it has any real benefit).

ricardoasmarques:
After thinking about this once again, I concluded that maybe we really need to specify the `bootstrap_minion`, but we need to be able to disable the bootstrap itself, something like:
```
if grains[id] == bootstrap_minion:
  if cephadm_bootstrap_enabled:
    run cephadm bootstrap
  else
    configure ssh orhestrator
```
I'll submit a proposal in the context of https://github.com/ceph/ceph-salt/issues/265 as soon as possible.

sebastian-philipp:
`s/configure cephdm orchestrator/` :smile:

ricardoasmarques:
@tserong I think this PR can be closed in favor of https://github.com/ceph/ceph-salt/pull/315

tserong:
Closing in favour of #315

PR #275, opened 2020-06-23
[DNM] Optionally deploy monitoring stack on bootstrap node

### LINKED ISSUE #274: Add parameter to control "--skip-monitoring-stack"
By default, unless "--skip-monitoring-stack" is given, "cephadm bootstrap" will deploy the monitoring stack (prometheus, grafana, alertmanager, node-exporter) automatically along with the first MON and MGR.

Currently, ceph-salt **always** gives "--skip-monitoring-stack", so there is no way for a ceph-salt user to take advantage of this feature of "cephadm bootstrap".

Let's add a config parameter to ceph-salt to give the user control over this option.

smithfarm:
d59c8eb9d1e96aa9d7424e0a236471019d84ae6e added "--skip-monitoring-stack"
to the "cephadm bootstrap" command, which had the effect of
unconditionally removing the user's access to this cephadm feature.

There might be scenarios where the user would want to deploy the
monitoring stack on the bootstrap node, so add a configurable for it.

Fixes: https://github.com/ceph/ceph-salt/issues/274

---

To Do:

- [ ] add configurables and other code necessary for including monitor container image paths in configshell "container->images->grafana"
- [ ] `cephadm bootstrap` supports specifying alternate monitoring container image paths

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/cephbootstrap.sls
@@ -44,7 +44,9 @@ run cephadm bootstrap:
 {%- endif %}
                 --output-keyring /etc/ceph/ceph.client.admin.keyring \
                 --output-config /etc/ceph/ceph.conf \
+{%- if not pillar['ceph-salt']['bootstrap_node_monitoring_stack'] %}
                 --skip-monitoring-stack \
+{%- endif %}
                 --skip-prepare-host \
                 --skip-pull \
                 --skip-ssh \
--- a/ceph_salt/config_shell.py
+++ b/ceph_salt/config_shell.py
@@ -497,6 +497,12 @@ add location=172.17.0.1:5000/docker.io prefix=docker.io insecure=true
                 'help': 'Bootstrap Mon IP',
                 'handler': PillarHandler('ceph-salt:bootstrap_mon_ip')
             },
+            'monitoring_stack': {
+                'type': 'flag',
+                'help': ('Whether to deploy the whole monitoring stack on the '
+                         'bootstrap node'),
+                'handler': PillarHandler('ceph-salt:bootstrap_node_monitoring_stack', False)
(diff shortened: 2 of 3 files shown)


smithfarm:
If I install ceph-salt from this PR and enable the new `bootstrap_node_monitoring_stack` configurable, I don't get any monitoring stack. I suspect it's because `ceph-salt` runs `cephadm bootstrap` with `--skip-pull`, so there are no monitoring container images in the system.

The ceph base image is passed to `cephadm bootstrap` using an environment variable:

```
        CEPHADM_IMAGE={{ pillar['ceph-salt']['container']['images']['ceph'] }} \
        cephadm --verbose bootstrap ....
```

Are there analogous environment variables for the "grafana", "node_exporter", "prometheus", and "alertmanager" container images?

smithfarm:
Added "blocked-by-cephadm" because I suspect that it might not (yet) be possible to specify alternate image paths on the `cephadm bootstrap` command line for all the monitoring-stack containers.

p-se:
> Added "blocked-by-cephadm" because I suspect that it might not (yet) be possible to specify alternate image paths on the `cephadm bootstrap` command line for all the monitoring-stack containers.

I wonder if it really is necessary to specify those alternative (downstream) container paths for monitoring components. `ceph-salt` depends on the packages that are available on the distribution it is run on, doesn't it? And if the distribution is from downstream, it should already contain a patched code with the correct URL to the downstream monitoring containers.

Relates to https://github.com/SUSE/ceph/pull/398

p-se:
Oh, I see the problem now. Not all of those paths/URLs are already available! For instance, Grafana will be available on the next milestone release, but the others are already available. Due to that, this might be a testing issue solely, but it'll resolve itself for production environments once the next release is out of the door.

smithfarm:
"--skip-monitoring-stack" is the right thing for ceph-salt to be doing here

PR #276, opened 2020-06-23
Optimize "ceph_orch.wait_for_admin_host" state

ricardoasmarques:
Deployment of large clusters is failing with the following error, which is caused by multiple `salt-call` executions in parallel on `admin` minions:

```
2020-06-23 12:32:13,238 [salt.minion      :2011][WARNING ][30636] The minion failed to return the job information for job 20200623123149042477. This is often due to the master being shut down or overloaded. If the master is running, consider increasing the worker_threads value.
```

This PR optimize `ceph_orch.wait_for_admin_host` state, which will fix the error above.

1) Do not use `salt-call` for better performance.

2) Also, use timeout to guarantee that `ceph orch status` will not hang.

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_orch.py
+++ b/ceph-salt-formula/salt/_modules/ceph_orch.py
@@ -2,14 +2,11 @@
 import json
 
 def configured():
-    ret = __salt__['cmd.run_all']("sh -c 'type ceph'")
-    if ret['retcode'] != 0:
-        return False
     if not __salt__['file.file_exists']("/etc/ceph/ceph.conf"):
         return False
     if not __salt__['file.file_exists']("/etc/ceph/ceph.client.admin.keyring"):
         return False
-    ret = __salt__['cmd.run_all']("ceph orch status")
+    ret = __salt__['cmd.run_all']("timeout 60 ceph orch status")
     if ret['retcode'] != 0:
         return False
     return True
--- a/ceph-salt-formula/salt/_states/ceph_orch.py
+++ b/ceph-salt-formula/salt/_states/ceph_orch.py
@@ -1,5 +1,4 @@
 # -*- encoding: utf-8 -*-
-import json
 import time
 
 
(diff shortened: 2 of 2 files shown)


ceph-salt-formula/salt/_modules/ceph_orch.py
```diff
@@ -2,14 +2,11 @@
...
 def configured():
-    ret = __salt__['cmd.run_all']("sh -c 'type ceph'")
-    if ret['retcode'] != 0:
-        return False
     if not __salt__['file.file_exists']("/etc/ceph/ceph.conf"):
         return False
     if not __salt__['file.file_exists']("/etc/ceph/ceph.client.admin.keyring"):
         return False
-    ret = __salt__['cmd.run_all']("ceph orch status")
+    ret = __salt__['cmd.run_all']("timeout 15 ceph orch status")
```
smithfarm: could we double the timeout value to 30 seconds? 15 seconds makes me nervous

ceph-salt-formula/salt/_states/ceph_orch.py
```diff
@@ -25,14 +24,14 @@ def wait_for_admin_host(name, timeout=1800):
             if provisioned:
                 configured_ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
                                                         "-i /tmp/ceph-salt-ssh-id_rsa root@{} "
-                                                        "'salt-call ceph_orch.configured "
-                                                        "--out=json --out-indent=-1'".format(
+                                                        "'if [[ -f /etc/ceph/ceph.conf "
+                                                        "&& -f /etc/ceph/ceph.client.admin.keyring ]]; "
+                                                        "then timeout 15 ceph orch status; "
```
smithfarm: Could we double the timeout to 30 seconds?

ceph-salt-formula/salt/_modules/ceph_orch.py
ricardoasmarques: Done

ceph-salt-formula/salt/_states/ceph_orch.py
ricardoasmarques: Done

smithfarm:
I'm testing this now, together with https://github.com/SUSE/sesdev/pull/363

smithfarm:
> I'm testing this now, together with [SUSE/sesdev#363](https://github.com/SUSE/sesdev/pull/363)

The test failed ~and the failure appears to be related to this PR~: https://gitlab.suse.de/snippets/701

UPDATE: Not sure if this is relevant, but `curl` is not installed in the environment:

```
master:~ # curl
If 'curl' is not a typo you can use command-not-found to lookup the package that contains it, like this:
    cnf curl
```

UPDATE: looks like a transient failure when "zypper install cephadm" was attempted

smithfarm (approved):

swiftgist:
Doesn't `ceph orch status` only return an in memory data structure when queried?  That is, no other remote operations happen when `ceph orch status` is run.  If so, under what conditions would a success take 35 seconds?  In the fresh deployments now, I believe `ceph orch status` easily returns within 5 seconds, so 15 seconds doesn't sound terribly unreasonable.

To put this another way, let's say that a 15 second timeout stopped a call that would have been successful.  Do you think it's unreasonable to have the admin investigate why the mgr is taking more than 15 seconds to respond rather than continue?

ricardoasmarques:
> Doesn't `ceph orch status` only return an in memory data structure when queried? That is, no other remote operations happen when `ceph orch status` is run. If so, under what conditions would a success take 35 seconds? In the fresh deployments now, I believe `ceph orch status` easily returns within 5 seconds, so 15 seconds doesn't sound terribly unreasonable.
> 
> To put this another way, let's say that a 15 second timeout stopped a call that would have been successful. Do you think it's unreasonable to have the admin investigate why the mgr is taking more than 15 seconds to respond rather than continue?

If `/etc/ceph/ceph.conf` is pointing to an invalid MON ip adddress, `ceph orch status` will hang forever. This was my main motivation to add a timeout here.

And yes, 15 seconds should be more than enough for `ceph orch status` to return, but let's keep pessimistic for now since it the safest option.

PR #283, opened 2020-07-01
Check 'ceph orch status' output

ricardoasmarques:
~~**After https://github.com/ceph/ceph/pull/35805 is backported (https://github.com/ceph/ceph/pull/35898)**~~

--- 

Instead of simply rely on return code, we need to check if "available: true" is returned, because:

```
master:~ # ceph orch status --format=json

{"available": true, "backend": "cephadm"}
master:~ # echo $?
0

master:~ # ceph cephadm clear-key

master:~ # ceph orch status --format=json

{"available": false, "backend": "cephadm", "reason": "SSH keys not set. Use `ceph cephadm set-priv-key` and `ceph cephadm set-pub-key` or `ceph cephadm generate-key`"}

master:~ # echo $?
0
```

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_orch.py
+++ b/ceph-salt-formula/salt/_modules/ceph_orch.py
@@ -6,10 +6,11 @@ def configured():
         return False
     if not __salt__['file.file_exists']("/etc/ceph/ceph.client.admin.keyring"):
         return False
-    ret = __salt__['cmd.run_all']("timeout 60 ceph orch status")
-    if ret['retcode'] != 0:
+    status_ret = __salt__['cmd.run_all']("timeout 60 ceph orch status --format=json")
+    if status_ret['retcode'] != 0:
         return False
-    return True
+    status = json.loads(status_ret['stdout'])
+    return status.get('available', False)
 
 def host_ls():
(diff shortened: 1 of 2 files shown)


ceph-salt-formula/salt/_modules/ceph_orch.py
```diff
@@ -6,8 +6,11 @@ def configured():
         return False
     if not __salt__['file.file_exists']("/etc/ceph/ceph.client.admin.keyring"):
         return False
-    ret = __salt__['cmd.run_all']("timeout 60 ceph orch status")
-    if ret['retcode'] != 0:
+    status_ret = __salt__['cmd.run_all']("timeout 60 ceph orch status --format=json")
+    if status_ret['retcode'] != 0:
+        return False
+    status = json.loads(status_ret['stdout'])
+    if not status.get('available'):
```
smithfarm: `status.get('available', False)` ?

smithfarm (approved):

ceph-salt-formula/salt/_modules/ceph_orch.py
ricardoasmarques: Done

PR #297, opened 2020-07-15
Set 'unqualified-search-registries' based on OS distro

ricardoasmarques:
For SUSE distro, `registry.suse.com` should be used instead of `docker.io`.

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
+++ b/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
@@ -3,7 +3,11 @@
 # For more information on this configuration file, see containers-registries.conf(5)
 
 # An array of host[:port] registries to try when pulling an unqualified image, in order
+{% if grains.get('os_family') == 'Suse' and grains.get('osfullname') == 'SLES' %}
+unqualified-search-registries = ["registry.suse.com"]
+{% else %}
 unqualified-search-registries = ["docker.io"]
+{% endif %}
 
 {% for reg in registries %}
 [[registry]]


smithfarm (dismissed):

ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
```diff
@@ -3,7 +3,11 @@
 # For more information on this configuration file, see containers-registries.conf(5)
 
 # An array of host[:port] registries to try when pulling an unqualified image, in order
+{% if grains['os_family'] == 'Suse' %}
+unqualified-search-registries = ["registry.suse.com"]
```
sebastian-philipp, line 7: what's `grains['os_family']` for opensuse?

ricardoasmarques: It's `Suse`:

```
master:~ # cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.2 Beta"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.2"
PRETTY_NAME="openSUSE Leap 15.2 Beta"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.2"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"

master:~ # salt-call grains.get os_family
local:
    Suse
```

sebastian-philipp: which would be wrong to point leap to registry.suse.com. hm. I guess we need to distinguish opensuse and sle here.

smithfarm: We could deal with that later, in a follow-up PR. This patch solves the immediate problem. (After all, this code path is not even triggered in the typical use case, right? Users are expected to specify a full container image path, not a relative one.)

smithfarm: > which would be wrong to point leap to registry.suse.com. hm. I guess we need to distinguish opensuse and sle here.

We could check the "os_family" first, and if it is "Suse", we could then look into the "osfinger" grain to distinguish between SLE and openSUSE. But it's fragile and I don't like it...

ricardoasmarques: @sebastian-philipp ok, SLE should point to `registry.suse.com`, but where should we point `Leap`? `docker.io`?

bk201 (approved):

ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
ricardoasmarques: I've now updated this PR to also check `osfullname`.

**Leap:**
```
# salt-call grains.get os_family
local:
    Suse

# salt-call grains.get osfullname
local:
    Leap
```

**SLES:**
```
# salt-call grains.get os_family
local:
    Suse

# salt-call grains.get osfullname
local:
    SLES
```

sebastian-philipp (approved):

p-se (approved):

smithfarm:
Please note that my "disagreement" here is not "strong disagreement". If the PR gets merged over my disagreement, I won't be offended or upset, and life will go on as before :-)

ricardoasmarques:
@smithfarm I've now pushed a new PR with your proposal of NOT supporting unqualified names - https://github.com/ceph/ceph-salt/pull/299

If everyone agrees with the new proposal I'll then close this one.

ricardoasmarques:
Closing in favor of https://github.com/ceph/ceph-salt/pull/299

PR #298, opened 2020-07-16
Only allow authentication on a single registry

ricardoasmarques:
Instead of allowing the user to login to multiple registries (which in practice is not useful), we should only support login to a single registry.

This will allow us to better integrate with `cephadm` (https://github.com/ceph/ceph/pull/36012) in a follow-up PR.


smithfarm (approved):
Looks like a nice simplification

PR #299, opened 2020-07-16
Drop unqualified image name support

ricardoasmarques:
**This is an alternative to https://github.com/ceph/ceph-salt/pull/297.**

By removing the `unqualified-search-registries` setting, we force the user to always specify image full name.

**Fail:**
```
# podman pull ses/7/ceph/ceph:15.2.3.455.3.373
Error: error pulling image "ses/7/ceph/ceph:15.2.3.455.3.373": unable to pull ses/7/ceph/ceph:15.2.3.455.3.373: image name provided is a short name and no search registries are defined in the registries config file.
```

**Works:**
```
# podman pull registry.suse.com/ses/7/ceph/ceph:15.2.3.455.3.373
Trying to pull registry.suse.com/ses/7/ceph/ceph:15.2.3.455.3.373...
Getting image source signatures
Copying blob df41305a59b1 done  
Copying blob e258f0c054bc done  
Copying config 4ea80648d1 done  
Writing manifest to image destination
Storing signatures
4ea80648d17bea19f033934375b0668378c77c7e33a04eca9802d09aa3ce9707
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,8 @@ In this case, it is up to the user to ensure that some form of time syncing is
 configured and running on all machines in the cluster before triggering
 `ceph-salt apply`. This is because `cephadm` will refuse to run without it.
 
-Finally we need to set the Ceph container image path:
+Finally we need to set the Ceph container image absolute path (relative path is
+not supported):
 
 ```
 /containers/images/ceph set docker.io/ceph/ceph:v15.2.2
--- a/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
+++ b/ceph-salt-formula/salt/ceph-salt/files/registries.conf.j2
@@ -2,9 +2,6 @@
 # {% include "ceph-salt/files/managed-header.txt.j2" ignore missing %}
 # For more information on this configuration file, see containers-registries.conf(5)
 
-# An array of host[:port] registries to try when pulling an unqualified image, in order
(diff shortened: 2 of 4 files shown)


smithfarm:
I see in the README.md it says:

```
Finally we need to set the Ceph container image path:

    /containers/images/ceph set docker.io/ceph/ceph:v15.2.2
```

Maybe just add a note here that only absolute container image paths are supported?

And possibly add a validation test to `/containers/images/ceph set` so that if the substring up to the first `/` does not contain at least one `.`,  an error is displayed? The error could say something like: "A relative image path was given, but only absolute image paths are supported."

smithfarm (approved):
The validation test could be implemented in a follow-on PR

bk201 (approved):

ricardoasmarques:
> I see in the README.md it says:
> 
> ```
> Finally we need to set the Ceph container image path:
> 
>     /containers/images/ceph set docker.io/ceph/ceph:v15.2.2
> ```
> 
> Maybe just add a note here that only absolute container image paths are supported?
> 
> And possibly add a validation test to `/containers/images/ceph set` so that if the substring up to the first `/` does not contain at least one `.`, an error is displayed? The error could say something like: "A relative image path was given, but only absolute image paths are supported."

Done

smithfarm (approved):

PR #306, opened 2020-07-28
Purge ceph clusters

### LINKED ISSUE #242: Purge the whole cluster
It should be possible to remove all daemons by running `ceph-salt purge` command.

We can remove all daemons by running `cephadm rm-cluster --fsid $FSID --force` on all cephadm minions.

ricardoasmarques:
~~**After https://github.com/ceph/ceph-salt/pull/307**~~

---

This PR adds a `ceph-salt purge` that can be used to purge/destroy/remove all ceph clusters, useful for POCs:

Screenshot from 2020-07-28 12-33-22

Fixes: https://github.com/ceph/ceph-salt/issues/242

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -1,6 +1,7 @@
 # -*- encoding: utf-8 -*-
 import json
 import socket
+import time
 
 def _send_event(tag, data):
     __salt__['event.send'](tag, data=data)
@@ -65,3 +66,10 @@ def probe_ntp(ahost):
         return 1
     except:
         return 3
+
+def is_safety_disengaged():
+    execution = __pillar__['ceph-salt'].get('execution', {})
+    safety_disengage_time = execution.get('safety_disengage_time')
+    if safety_disengage_time and safety_disengage_time + 60 > time.time():
+        return True
+    return False
--- a/ceph-salt-formula/salt/ceph-salt/purge/init.sls
+++ b/ceph-salt-formula/salt/ceph-salt/purge/init.sls
@@ -1,0 +1,16 @@
+{% if 'ceph-salt' in grains and grains['ceph-salt']['member'] %}
+
+check safety:
+   ceph_salt.check_safety:
+     - failhard: True
+
+remove clusters:
+   ceph_orch.rm_clusters:
+     - failhard: True
+
+{% else %}
+
+nothing to do in this node:
+  test.nop
+
+{% endif %}
--- a/ceph_salt/__init__.py
+++ b/ceph_salt/__init__.py
@@ -11,7 +11,7 @@ from .config_shell import run_config_cmdline, run_config_shell, run_status, run_
 from .exceptions import CephSaltException
 from .logging_utils import LoggingUtil
 from .terminal_utils import check_root_privileges, PrettyPrinter as PP
-from .execute import CephSaltExecutor
+from .execute import CephSaltExecutor, run_disengage_safety, run_purge
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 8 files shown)


smithfarm:
I'm confused about the plural "all Ceph clusters". There can be more than one?

ricardoasmarques:
> I'm confused about the plural "all Ceph clusters". There can be more than one?

If there are more than one (manually deployed using cephadm directly), this PR will remove all of them.

smithfarm:
I suggest to implement the equivalent of DeepSea's "disengage safety" to protect users from shooting themselves with this.

I found "disengage safety" to be a useful feature, similar to mon_allow_pool_delete which must be set before "ceph osd pool delete" will work. It would be a shame for our users to lose this feature just because we implement a new tool.

ricardoasmarques:
> I suggest to implement the equivalent of DeepSea's "disengage safety" to protect users from shooting themselves with this.
> 
> I found "disengage safety" to be a useful feature, similar to mon_allow_pool_delete which must be set before "ceph osd pool delete" will work. It would be a shame for our users to lose this feature just because we implement a new tool.

I've now pushed a `ceph-salt disengage-safety` command that will disable safery during `60` seconds.

ceph-salt-formula/salt/ceph-salt/purge/init.sls
```diff
@@ -0,0 +1,16 @@
+{% if 'ceph-salt' in grains and grains['ceph-salt']['member'] %}
+
+ceck safety:
```
smithfarm: s/ceck/check/

ricardoasmarques: Fixed

ceph-salt-formula/salt/_modules/ceph_salt.py
```diff
@@ -65,3 +66,10 @@ def probe_ntp(ahost):
         return 1
     except:
         return 3
+
+def is_safety_disengaged():
+    execution = __pillar__['ceph-salt'].get('execution', {})
+    safety_disengage_time = execution.get('safety_disengage_time')
+    if not safety_disengage_time or safety_disengage_time + 60 <= time.time():
+        return False
+    return True
```
smithfarm: could we reverse the logic here to make it easier for my poor brain to comprehend?

```
    if safety_disengage_time and safety_disengage_time + 60 > time.time():
        return True
    return False
```

This makes it more obvious that the function defaults to returning "False" and only returns True under certain strict conditions.

ceph_salt/__init__.py
```diff
@@ -107,5 +108,39 @@ def apply(non_interactive, minion_id):
...
 
 
+def _prompt_proceed(msg, default):
+    really_want_to = click.prompt(
+        '{} (y/n):'.format(msg),
+        type=str,
+        default=default,
+    )
+    if really_want_to.lower()[0] != 'y':
+        raise click.Abort()
```
smithfarm, line 117: :+1: :-)

ceph-salt-formula/salt/_modules/ceph_salt.py
ricardoasmarques: Done

ricardoasmarques:
> > I'm confused about the plural "all Ceph clusters". There can be more than one?
> 
> If there are more than one (manually deployed using cephadm directly), this PR will remove all of them.

@smithfarm I've now updated this PR to only purge one cluster (the one reported by `ceph -s`).

ricardoasmarques:
For better usability, I've also updated this PR to fail before-hand when safety is enabled (before start running `salt state.apply`):

Screenshot from 2020-07-30 10-30-55

smithfarm (approved):

sebastian-philipp (approved):

PR #303, opened 2020-07-24
Add 'ceph-salt update [--reboot] [minion_id]' command

ricardoasmarques:
~~**After https://github.com/ceph/ceph-salt/pull/307**~~
~~**After https://github.com/ceph/ceph/pull/36232 is backported (https://github.com/ceph/ceph/pull/36450)**~~

---

This PR introduces the support for orchestrated updates and reboot, by providing a new `ceph-salt update [--reboot] [minion_id]` command to **update** all minions in **parallel**, and **reboot** them **sequentially** if needed.

By default `ceph-salt update` comand will update all hosts, in parallel, without rebooting them.
- If `--reboot` option is provided, ceph-salt will reboot minions sequentially (if needed).
- If `minion_id` is provided, update/reboot will only be applied on that minion.

Screenshot from 2020-07-24 10-28-35

(Note that this PR is moving a lot of files in order to make it possible to implement new commands)

Signed-off-by: Ricardo Marques <rimarques@suse.com>

PR #307, opened 2020-07-28
Refactor to support adding new salt state calls

ricardoasmarques:
This PR do a refactor in order to make it possible to call new salt states from `ceph-salt` tool.

ATM, it's only possible to call `ceph-salt` state using `ceph-salt apply` command, but in the future, we will need more commands, e.g:

- `ceph-salt apply` should call `state.apply ceph-salt`
- `ceph-salt update` should call `state.apply ceph-salt.update`
- `ceph-salt purge` should call `state.apply ceph-salt.purge`

This PR is also a preparation to deal with sequential operations (e.g. system reboot, cluster shutdown).

It also adds support to trigger `state_begin`, `state_end`, `step_begin` and `step_end` event from the `ceph-salt` module.

**Note:** These changes will be used in https://github.com/ceph/ceph-salt/pull/303 and https://github.com/ceph/ceph-salt/pull/306.


smithfarm (approved):

smithfarm:
I already looked at this when it was part of #303 - I'm fine with it.

PR #308, opened 2020-07-28
cephbootstrap: run bootstrap with --apply-spec

smithfarm:
As reported in the downstream bug cited below, "cephadm bootstrap"
is now deploying 5 MONs and 2 MGRs by default.

The expectation is that "ceph-salt apply" will cause one (and only one)
MON and one (and only one) MGR to be deployed, and that both the MON and
the MGR will be deployed on the node specified by the user as the
bootstrap node.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1174422

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
@@ -5,15 +5,16 @@
 {{ macros.begin_stage('Bootstrap the Ceph cluster') }}
 
 {% set bootstrap_ceph_conf = pillar['ceph-salt'].get('bootstrap_ceph_conf', {}) %}
+{% set bootstrap_ceph_conf_tmpfile = "/tmp/bootstrap-ceph.conf" %}
 
 create bootstrap ceph conf:
   cmd.run:
     - name: |
-        echo -en "" > /tmp/bootstrap-ceph.conf
+        echo -en "" > {{ bootstrap_ceph_conf_tmpfile }}
 {% for section, settings in bootstrap_ceph_conf.items() %}
-        echo -en "[{{ section }}]\n" >> /tmp/bootstrap-ceph.conf
+        echo -en "[{{ section }}]\n" >> {{ bootstrap_ceph_conf_tmpfile }}
 {% for setting, value in settings.items() %}
-        echo -en "        {{ setting }} = {{ value }}\n" >> /tmp/bootstrap-ceph.conf
+        echo -en "        {{ setting }} = {{ value }}\n" >> {{ bootstrap_ceph_conf_tmpfile }}
 {% endfor %}
 {% endfor %}
     - failhard: True
@@ -37,7 +38,7 @@ run cephadm bootstrap:
     - name: |
         CEPHADM_IMAGE={{ pillar['ceph-salt']['container']['images']['ceph'] }} \
         cephadm --verbose bootstrap \
-                --config /tmp/bootstrap-ceph.conf \
+                --config {{ bootstrap_ceph_conf_tmpfile }} \
 {%- if not pillar['ceph-salt']['dashboard']['password_update_required'] %}
                 --dashboard-password-noupdate \
 {%- endif %}
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
@@ -6,6 +6,23 @@
 
 {% set bootstrap_ceph_conf = pillar['ceph-salt'].get('bootstrap_ceph_conf', {}) %}
 {% set bootstrap_ceph_conf_tmpfile = "/tmp/bootstrap-ceph.conf" %}
+{% set bootstrap_spec_yaml_tmpfile = "/tmp/bootstrap-spec.yaml" %}
+
+create static bootstrap yaml:
+  cmd.run:
+    - name: |
+        echo -en "" > {{ bootstrap_spec_yaml_tmpfile }}
+        echo -en "service_type: mgr\n" >> {{ bootstrap_spec_yaml_tmpfile }}
+        echo -en "service_name: mgr\n" >> {{ bootstrap_spec_yaml_tmpfile }}
+        echo -en "placement:\n" >> {{ bootstrap_spec_yaml_tmpfile }}
+        echo -en "    hosts:\n" >> {{ bootstrap_spec_yaml_tmpfile }}
+        echo -en "        - {{ grains['host'] }}\n" >> {{ bootstrap_spec_yaml_tmpfile }}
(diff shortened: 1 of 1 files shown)


smithfarm:
c3bf1ebf4c07b6f3ec6a2da671a3adc590c8f9f4 started installing "ceph-base" RPM on minions before the RPM was added to the product Media: added temporary revert for testing

ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
```diff
@@ -6,6 +6,22 @@
...
 
+create static bootstrap yaml:
+  cmd.run:
+    - name: |
+        echo -en "" > /tmp/static_bootstrap.yaml
+        echo -en "service_type: mgr\n" >> /tmp/static_bootstrap.yaml
+        echo -en "service_name: mgr\n" >> /tmp/static_bootstrap.yaml
+        echo -en "placement:\n" >> /tmp/static_bootstrap.yaml
+        echo -en "    hosts:\n" >> /tmp/static_bootstrap.yaml
+        echo -en "        - {{ pillar['ceph-salt']['bootstrap_node_hostname'] }}\n" >> /tmp/static_bootstrap.yaml
```
ricardoasmarques: Instead of adding a new pillar item, can we simply read `host` grain like we are doing in https://github.com/ceph/ceph-salt/blob/master/ceph-salt-formula/salt/ceph-salt/apply/cephorch.sls#L8 ?

Note that this is running in the bootstrap minion, so it should work.

ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
```diff
@@ -6,6 +6,22 @@
...
+        echo -en "---\n" >> /tmp/static_bootstrap.yaml
+        echo -en "service_type: mon\n" >> /tmp/static_bootstrap.yaml
+        echo -en "service_name: mon\n" >> /tmp/static_bootstrap.yaml
+        echo -en "placement:\n" >> /tmp/static_bootstrap.yaml
+        echo -en "    hosts:\n" >> /tmp/static_bootstrap.yaml
+        echo -en "        - {{ pillar['ceph-salt']['bootstrap_node_hostname'] }}\n" >> /tmp/static_bootstrap.yaml
```
ricardoasmarques: Perhaps `{{ grains['host'] }}`?

ceph_salt/config_shell.py
```diff
@@ -96,6 +96,7 @@ def save(self, value):
             raise MinionDoesNotExistInConfiguration(value)
         node = CephNodeManager.ceph_salt_nodes()[value]
         PillarManager.set('ceph-salt:bootstrap_mon_ip', node.public_ip)
+        PillarManager.set('ceph-salt:bootstrap_node_hostname', node.hostname)
```
ricardoasmarques: Hopefully, we can remove this.

sebastian-philipp:
lgtm

smithfarm:
So, I added a commit to specify the Mon IP in the placement YAML. It didn't make sense to me to specify the same piece of information in two places, so I dropped the `--mon-ip` option. However, when invoked without `--mon-ip` or `--mon-addrv`, `cephadm bootstrap` barfs out a Traceback:

```
Traceback (most recent call last):
  File "/usr/sbin/cephadm", line 5046, in <module>
    r = args.func()
  File "/usr/sbin/cephadm", line 1214, in _default_image
    return func()
  File "/usr/sbin/cephadm", line 2537, in command_bootstrap
    raise Error('must specify --mon-ip or --mon-addrv')
__main__.Error: must specify --mon-ip or --mon-addrv
```

Opened https://tracker.ceph.com/issues/46782 to track this upstream.

smithfarm:
@ricardoasmarques I just tested this again and it works for me.

ricardoasmarques:
> So, I added a commit to specify the Mon IP in the placement YAML. It didn't make sense to me to specify the same piece of information in two places, so I dropped the `--mon-ip` option. However, when invoked without `--mon-ip` or `--mon-addrv`, `cephadm bootstrap` barfs out a Traceback:
> 
> ```
> Traceback (most recent call last):
>   File "/usr/sbin/cephadm", line 5046, in <module>
>     r = args.func()
>   File "/usr/sbin/cephadm", line 1214, in _default_image
>     return func()
>   File "/usr/sbin/cephadm", line 2537, in command_bootstrap
>     raise Error('must specify --mon-ip or --mon-addrv')
> __main__.Error: must specify --mon-ip or --mon-addrv
> ```
> 
> Opened https://tracker.ceph.com/issues/46782 to track this upstream.

@sebastian-philipp is your approval still valid now that we are specifying the Mon IP in the placement YAML?

ricardoasmarques (approved):
lgtm

sebastian-philipp:
we have to specify the `--mon-ip` here. bootstrap works like so:

1. create the first mon using `--mon-ip`
1. create the first mgr
1. configure dashboard etc...
1. `ceph orch apply -i spec.yaml`
1. `mgr/cephadm` will follow the instructions that were passed in the spec.

Which means, we're looking at the spec **much** later. Meaning after bootstrapping the first mon. There is no way for now to look into the spec.yaml at this point and extract the mon ip.

sebastian-philipp:
lgtm

smithfarm:
@sebastian-philipp Ah! Now I understand what is missing. If we had a `--no-apply-spec` option, we could restore the original behavior of `cephadm bootstrap`, which was to create 1 MON and 1 MGR on the node where it is running.

Then we wouldn't need to bend over backwards sending `cephadm bootstrap` a spec file which merely states what has already done, in order to prevent `cephadm bootstrap` from "helpfully" creating 5 MONs and 2 MGRs.

What do you think?

smithfarm:
Opened https://tracker.ceph.com/issues/46801 to track the issue, so we don't block this PR.

sebastian-philipp:
> @sebastian-philipp Ah! Now I understand what is missing. If we had a `--no-apply-spec` option, we could restore the original behavior of `cephadm bootstrap`, which was to create 1 MON and 1 MGR on the node where it is running.

The `--no-apply-spec` is called `--orphan-initial-daemons`, but you really don't want to do that, except if you're Teuthology, as it leaves the deplyed daemons orphaned and scheduled for potential removal


> 
> Then we wouldn't need to bend over backwards sending `cephadm bootstrap` a spec file which merely states what has already done, in order to prevent `cephadm bootstrap` from "helpfully" creating 5 MONs and 2 MGRs.
> 
> What do you think?

smithfarm:
> The `--no-apply-spec` is called `--orphan-initial-daemons`, but you really don't want to do that, except if you're Teuthology, as it leaves the deplyed daemons orphaned and scheduled for potential removal

In that case, it sounds like we need a "--no-apply-spec" that is just like `--orphan-initial-daemons`, except it doesn't orphan the initial daemons :-)

PR #311, opened 2020-07-29
chrony.conf: omit useless and counterproductive options

### LINKED ISSUE #310: local stratum 10 and manual in chrony.conf - why?
We should not rely on local time and I am not sure why "manual" is in the template.

smithfarm:
Fixes: https://github.com/ceph/ceph-salt/issues/310


ricardoasmarques (approved):
Tested and still works.

PR #315, opened 2020-08-05
Make bootstrap minion optional

### LINKED ISSUE #265: Make bootstrap_mon optional

ricardoasmarques:
For considering upgrades where a Ceph cluster already exists, bootstrap minion should not be required.

1. Bootstrap minion is no longer required if `ceph -s` command is already available on any admin minion
2. Ceph-salt will now explicitly enable `cephadm` module and set orchestrator backend, even if the cluster is already bootstrapped
3. Ceph-salt will guarantee that SSH keys and SSH user are set

Fixes: https://github.com/ceph/ceph-salt/issues/265
Relates to: https://github.com/ceph/ceph-salt/pull/267
Relates to: https://github.com/SUSE/DeepSea/pull/1840


tserong (approved):
With the disclaimer that I haven't yet managed to test this myself, I think it looks OK :-)

What's up with the CI failure? Is that due to this PR, or something unrelated?

ricardoasmarques:
CI failure is unrelated

PR #321, opened 2020-08-06
Configure 'qualified-search-registries = ["docker.io"]'

### LINKED ISSUE #309: registries.conf should also have unqualified-search-registries = ["docker.io"]
registries.conf should also have unqualified-search-registries = ["docker.io"]

(just in case other containers are used later on..)

ricardoasmarques:
In https://github.com/ceph/ceph-salt/pull/299 we removed `qualified-search-registries = ["docker.io"]` from `registries.conf` but this is useful in case other containers are used later on.

However, we keep the restriction introduced in https://github.com/ceph/ceph-salt/pull/299 that forces the user to specify `ceph` image full name.

Fixes: https://github.com/ceph/ceph-salt/issues/309


sebastian-philipp (approved):

sebastian-philipp:
@p-se with this PR, is ceph/ceph#36315 still required?

bk201 (approved):

PR #322, opened 2020-08-10
'cephadm' MGR module should only be enabled after cluster bootstrapped

ricardoasmarques:
`cephadm` MGR module should be enabled by the bootstrap minion, or the first admin minion (if the bootstrap minion is not specified).

This will guarantee that `cephadm` MGR module will only be enabled after cluster is bootstrapped, to prevent the following error:

Screenshot from 2020-08-10 10-16-02


tserong (approved):

ricardoasmarques:
With this PR I wasn't able to reproduce the issue anymore.

PR #329, opened 2020-08-21
Port net.ping/net.iperf to ceph-salt as network.ping/network.iperf

swiftgist:
Keep previous history and make necessary changes to work in a ceph-salt enviroment.


swiftgist:
This runner and module does depend on the Salt master finding the runner.  I have not tracked down the package `salt-standalone-formulas-configuration-3000-2.9.x86_64` which contains `/etc/salt/master.d/standalone-formulas-configuration.conf`.  That configuration file should contain

```
file_roots:
  base:
    - /usr/share/salt-formulas/states
    - /srv/salt

runner_dirs:
  - /usr/share/salt-formulas/states/_runners
```

ricardoasmarques:
> This runner and module does depend on the Salt master finding the runner. I have not tracked down the package `salt-standalone-formulas-configuration-3000-2.9.x86_64` which contains `/etc/salt/master.d/standalone-formulas-configuration.conf`. That configuration file should contain
> 
> ```
> file_roots:
>   base:
>     - /usr/share/salt-formulas/states
>     - /srv/salt
> 
> runner_dirs:
>   - /usr/share/salt-formulas/states/_runners
> ```

Did you tried `salt-run saltutil.sync_runners` to make sure Master see new runners?

swiftgist:
> Did you tried `salt-run saltutil.sync_runners` to make sure Master see new runners?

I only have `salt '*' saltutil.sync_all` ingrained in my memory.  I didn't know they made a runner of the same name to confuse as many as possible.  I gave `salt-run saltutil.sync_all` and that does show the runner getting copied.  

If you want to test before running `ceph-salt apply`, then iperf needs to be installed and the runners and modules need to be sync'd.  I expect this to be troublesome in deployments.  For instance, these are the steps if the first thing an admin wants to verify is that the hardware seems configured properly.

```
# zypper in ceph-salt
# salt-run saltutil.sync_all 
# salt '*' cmd.run "zypper --non-interactive in iperf"
# salt-run minion.ping
```

The saltutil runner does copy the modules as well, it seems.  Also, every upgrade of ceph-salt will require the `sync_all`.  It's caused complaints for DeepSea.  Users normally think "I have installed the rpm", therefore "the software is ready".  With Salt, that's not the case.  And those new to Salt simply get frustrated out of the gate.

If the other subcommand `update` is going to be a prerequisite for `apply`, then maybe it could do the `sync_all` as well.

ceph-salt-formula/salt/ceph-salt/apply/software.sls
```diff
@@ -6,6 +6,7 @@ install required packages:
   pkg.installed:
     - pkgs:
       - iputils
+      - iperf
```
ricardoasmarques, line 8: I know this will be needed in minions, but for consistency, can we also add this to the spec file just to make the requirement more explicit? 
(We are already doing the same for other dependencies)

ceph-salt-formula/salt/_modules/multi.py
```diff
@@ -0,0 +1,275 @@
...
+    Use iperf to test minion to server
+
+    CLI Example:
+    .. code-block:: bash
+    salt 'node' multi.iperf_server_cmd <server_name/ip>
+        cpu=<which_cpu_core default 0> port=<default 5200>
+    '''
+    if IPERF_PATH is None:
+        return LOCALHOST_NAME + ": iperf3 not found in path, please install"
+    iperf_cmd = ["/usr/bin/iperf3", "-s", "-D", "-A"+str(cpu), "-p"+str(port)]
```
ricardoasmarques, line 187: Can we do something like `salt-call pkg.install iperf` here, if iperf is not installed yet?

This will guarantee that `salt-run minion.iperf` will work even if `ceph-salt apply` was not executed yet.

ricardoasmarques:
What about renaming `minion.py` and `multi.py` to something more meaningful?

I imagine something like `profile.py` or `profiling.py` since this seems to be a `profiling` tool, or maybe `diagnostic.py` / `diagnose.py`is even better?

smithfarm:
> What about renaming `minion.py` and `multi.py` to something more meaningful?
> 
> I imagine something like `profile.py` or `profiling.py` since this seems to be a `profiling` tool, or maybe `diagnostic.py` is even better?

Good idea. My vote would be for `diagnostics.py`.

ricardoasmarques:
> If you want to test before running `ceph-salt apply`, then iperf needs to be installed and the runners and modules need to be sync'd. I expect this to be troublesome in deployments. For instance, these are the steps if the first thing an admin wants to verify is that the hardware seems configured properly.
> 
> ```
> # zypper in ceph-salt
> # salt-run saltutil.sync_all 
> # salt '*' cmd.run "zypper --non-interactive in iperf"
> # salt-run minion.ping
> ```

One possible solution is to migrate this runner to a `ceph-salt` command, e.g.:

```

@cli.group()
def diagnostic():
    pass


@diagnostic.command()
def ping():
    """
    Update all packages
    """
    run_ping()

@diagnostic.command()
def iperf():
    """
    Update all packages
    """
    run_iperf()
```

and let `ceph-salt` perform the sync and install required dependencies for you.

But this requires more refactoring to the runner, so we don't have to do it now, in this PR.

ricardoasmarques (approved):
lgtm

PR #332, opened 2020-08-24
execute: check for running jobs when failing to re-apply formula

### LINKED ISSUE #327: In theory, "ceph-salt apply" should be idempotent, but in practice there are issues
It seems fairly easy to get a cluster into the following state by re-running "ceph-salt apply" and interrupting it while it's still running:

```
# ceph-salt apply
```

Now, I understand I'm not supposed to interrupt it, and "ceph-salt apply" tries to make it difficult for me to do so, but we cannot just say: "don't interrupt it while it's running". What if the user experiences a power failure while it's running?

Also, it's not clear why this error message is displayed in this case. The ceph-salt-formula package is installed (of course).

Worse yet, it's not clear how to recover. I've heard of people completely re-imaging their bare metal nodes in order to re-run "ceph-salt apply" when something goes wrong. There should be a better way?

smithfarm:
Originally, we thought that the only case when

    salt -G ceph-salt:member state.sls_exists ceph-salt

can fail is when the ceph-salt-formula package is not installed. Now we
know it can also fail when jobs initiated by an earlier invocation of
the Salt Formula are still running.

Check for such jobs and display a different error message in this case.

Fixes: https://github.com/ceph/ceph-salt/issues/327

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -222,8 +222,7 @@ min-similarity-lines=4
 
 
 [MISCELLANEOUS]
-notes=FIXME,
-      XXX,
+notes=XXX,
       TODO
 
 [IMPORTS]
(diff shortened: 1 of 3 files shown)


ricardoasmarques (approved):
lgtm

Issue #148, opened 2020-03-27
ceph-salt "tox -e lint" broken for python 3.8


toabctl:
When I try to run `tox -elint` on openSUSE Tumbleweed with python 3.8, I get:

```
$ tox -relint
lint create: /home/tom/devel/ceph/ceph-salt/.tox/lint
lint installdeps: -r/home/tom/devel/ceph/ceph-salt/requirements.txt
ERROR: invocation failed (exit code 1), logfile: /home/tom/devel/ceph/ceph-salt/.tox/lint/log/lint-1.log
========================================================================= log start ==========================================================================
Processing /home/tom/devel/ceph/ceph-salt
Collecting mock==3.0.5
  Using cached https://files.pythonhosted.org/packages/05/d2/f94e68be6b17f46d2c353564da56e6fb89ef09faeeff3313a046cb810ca9/mock-3.0.5-py2.py3-none-any.whl
Collecting pycodestyle==2.5.0
  Using cached https://files.pythonhosted.org/packages/0e/0c/04a353e104d2f324f8ee5f4b32012618c1c86dd79e52a433b64fceed511b/pycodestyle-2.5.0-py2.py3-none-any.whl
Collecting pyfakefs==3.7
  Using cached https://files.pythonhosted.org/packages/31/5a/bbf8a22ee586482a339712b9b43f1df1a4c17d76a06384a27dac7b477065/pyfakefs-3.7-py2.py3-none-any.whl
Collecting pylint==2.4.4
  Using cached https://files.pythonhosted.org/packages/e9/59/43fc36c5ee316bb9aeb7cf5329cdbdca89e5749c34d5602753827c0aa2dc/pylint-2.4.4-py3-none-any.whl
Collecting pytest==5.3.1
  Using cached https://files.pythonhosted.org/packages/da/ed/d22d7f06eb1107271694ed2171b9d52e8eea38d9757124e75ba13324ac77/pytest-5.3.1-py3-none-any.whl
Collecting pytest-cov==2.8.1
  Using cached https://files.pythonhosted.org/packages/b9/54/3673ee8be482f81527678ac894276223b9814bb7262e4f730469bb7bf70e/pytest_cov-2.8.1-py2.py3-none-any.whl
Collecting pytest-runner==5.2
  Using cached https://files.pythonhosted.org/packages/16/45/81b5262c0efc08882bdf183b788e6d28e3d684863990996d8b60967d48da/pytest_runner-5.2-py2.py3-none-any.whl
[...]

PR #338, opened 2020-08-24
tox,travis: test on Python 3.8 instead of 3.7

### LINKED ISSUE #336: Python 3.8 has been released, so Travis should test against it instead of 3.7
Travis is currently configured to run unit tests in two environments: Python 3.6 and 3.7.

But now that Python 3.8 is out, we should change 3.7 to 3.8.

smithfarm:
And for local testing we can get away with 3.6 only. (It's convenient to
be able to run "tox" and not get all tests repeated in 3.8 after they
all ran once in 3.6.)

Fixes: https://github.com/ceph/ceph-salt/issues/336

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ matrix:
   include:
   - python: 3.6
     env: TOX_ENV=py36
-  - python: 3.7
-    env: TOX_ENV=py37
-  - python: 3.7
+  - python: 3.8
+    env: TOX_ENV=py38
+  - python: 3.8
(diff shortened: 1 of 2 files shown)


ricardoasmarques:
Relates to https://github.com/ceph/ceph-salt/issues/148

smithfarm:
I think this PR could be seen as being "the fix" for #148. (At least, it proves that #148 has been fixed in the meantime.)

ricardoasmarques (approved):
lgtm

PR #343, opened 2020-08-25
_send_event: write INFO message to minion log

### LINKED ISSUE #342: When steps and stages begin or end, write log message to the Salt minion log
Execution modules can write messages to the minion logs. Quoting from https://docs.saltstack.com/en/latest/ref/modules/index.html#log-output

```
Log Output

You can call the logger from custom modules to write messages to the minion logs. The following code snippet demonstrates writing log messages:

import logging

log = logging.getLogger(__name__)

log.info("Here is Some Information")
log.warning("You Should Not Do That")
log.error("It Is Busted")
```

Stage/step begin/end is implemented using a macro.

It looks like the macros are implemented using modules.

So adding a `log.info("...")` to the relevant module should be feasible?

smithfarm:
When states/steps begin or end, it will be useful to see that in the
minion log.

Fixes: https://github.com/ceph/ceph-salt/issues/342

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -1,10 +1,13 @@
 # -*- encoding: utf-8 -*-
 import json
+import logging
 import socket
 import time
 
+log = logging.getLogger(__name__)
 
 def _send_event(tag, data):
+    log.info("{}, data={}".format(tag, data))
     __salt__['event.send'](tag, data=data)
     return {
         'name': tag,


smithfarm:
doesn't do anything, and is not needed. See my comments in #342

ceph-salt-formula/salt/_modules/ceph_salt.py
```diff
@@ -1,10 +1,13 @@
 # -*- encoding: utf-8 -*-
 import json
+import logging
 import socket
 import time
 
+log = logging.getLogger(__name__)
 
 def _send_event(tag, data):
+    log.info("{}, data={}".format(tag, data))
```
ricardoasmarques, line 10: You need to do the same in `ceph-salt-formula/salt/_states/ceph_salt.py` file

PR #344, opened 2020-08-26
execute: write INFO message to log when stage/step begins/ends

smithfarm:


ricardoasmarques (changes requested):
We also need to log `self.minion`, otherwise, we don't know which minion triggered the event.

smithfarm:
test this please

ricardoasmarques (approved):
lgtm

PR #345, opened 2020-08-26
Remove '/system_update' config

### LINKED ISSUE #326: Now that we have "ceph-salt update [--reboot]" we should remove package update and reboot steps from "ceph-salt apply"
Currently, in the default case, "ceph-salt apply" will

1. update all packages on each node/minion (including the master node)
2. reboot all nodes except the master
3. continue on its way

This is a recipe for trouble, because:

1. the minions might not all come back from reboot
2. the master will have "processes using deleted files" (potentially lots of them, depending on how out-of-date the system was)

Both of these conditions (minions not returning from reboot and processes using deleted files) can cause the deployment to fail in difficult-to-debug ways.

Now that we have the `ceph-salt update [--reboot]` feature as a separate command, IMO it would make sense to:

1. rip out the software update and node reboot parts of "ceph-salt apply" (provision.sls)
2. document that the user is supposed to complete the following steps before running "ceph-salt apply":
    a. run `ceph-salt update --reboot` and also manually reboot the Salt Master node
    b. check that all systems are up-to-date (`salt \* cmd.run zypper -n up` shows "Nothing to do")
    c. check that all minions are responding, i.e.: `salt \* test.ping` shows all minions (that are expected to be present) have returned from reboot and have rejoined the Salt cluster

ricardoasmarques:
With this PR, `ceph-salt apply` will no longer perform system **updates** and **reboots**.

Updates must be done via `ceph-salt update` command which was now improved to support updating hosts when no ceph cluster is running yet.

Note that if no ceph cluster exists, ceph-salt will reboot hosts in parallel, otherwise reboots are orchestrated/sequential.

Fixes https://github.com/ceph/ceph-salt/issues/326


ceph-salt-formula/salt/ceph-salt/apply/cephconfigure.sls
```diff
@@ -2,14 +2,10 @@
 
 {% if 'cephadm' in grains['ceph-salt']['roles'] or 'admin' in grains['ceph-salt']['roles'] %}
 
-{{ macros.begin_stage('Find an admin host') }}
-
```
smithfarm, line 6: What is the rationale for dropping the "Find an admin host" stage begin and end messages?

ceph-salt-formula/salt/ceph-salt/apply/cephconfigure.sls
```diff
@@ -2,14 +2,10 @@
...
 find an admin host:
   ceph_orch.set_admin_host:
     - failhard: True
 
-{{ macros.end_stage('Find an admin host') }}
-
```
smithfarm, line 12: I'm not sure I agree to drop the "Find an admin host" begin_stage and end_stage macro calls . . . . ceph-salt might spend a lot of time (up to 1800 seconds) in this loop and without the begin_stage message it will be less clear what the minions are doing.

ceph-salt-formula/salt/ceph-salt/apply/cephconfigure.sls
ricardoasmarques: It's not being dropped.. it was moved to `ceph_orch.set_admin_host` state [1] because we don't want to show that message if a reboot is not needed [2].

Now you may be asking "why not using a Jinja `if` condition" for that?
Answer is... because `ceph-salt:execution:reboot_needed` is set dinamically in [3] and state file is renderer when you start `state.apply` (and will not be renderer again even if you change the value of grains used in `if` conditions during runtime).


[1] https://github.com/ceph/ceph-salt/pull/345/files/88380c09fee8dd750f8d4c20b183d4d9a9cb901f#diff-dbf4c37a5be102b76d21bbb73d262c20R16

[2] https://github.com/ceph/ceph-salt/pull/345/files/88380c09fee8dd750f8d4c20b183d4d9a9cb901f#diff-1a2c1f3cc1d8d46ea1184ada6d3a448bR35

[3] https://github.com/ceph/ceph-salt/pull/345/files/88380c09fee8dd750f8d4c20b183d4d9a9cb901f#diff-1a2c1f3cc1d8d46ea1184ada6d3a448bR25

ceph-salt-formula/salt/ceph-salt/apply/cephconfigure.sls
ricardoasmarques: Explained in my previous comment.

smithfarm (approved):

PR #353, opened 2020-09-01
Only pull ceph image on bootstrap minion

ricardoasmarques:
With this PR, ceph image will only be pulled on the bootstrap minion.

All other minions will pull ceph image via `cephadm` latter when needed.

The main motivation for this PR is to prevent `network overload` when pulling the image on all hosts in parallel (e.g. > 100 nodes)

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
@@ -2,6 +2,57 @@
 
 {% if grains['id'] == pillar['ceph-salt'].get('bootstrap_minion') %}
 
+{{ macros.begin_stage('Prepare to bootstrap the Ceph cluster') }}
+
+/var/log/ceph:
+  file.directory:
+    - user: ceph
+    - group: ceph
+    - mode: '0770'
+    - makedirs: True
+    - failhard: True
+
+{% set auth = pillar['ceph-salt'].get('container', {}).get('auth', {}) %}
+
+{% if auth %}
+
+{{ macros.begin_step('Login into registry') }}
+
+create ceph-salt-registry-json:
+  file.managed:
+    - name: /tmp/ceph-salt-registry-json
+    - source:
+        - salt://ceph-salt/files/registry-login-json.j2
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: '0600'
+    - backup: minion
+    - failhard: True
+
+login into registry:
+  cmd.run:
+    - name: |
+        cephadm registry-login \
+        --registry-url {{ auth.get('registry') }} \
+        --registry-username {{ auth.get('username') }} \
+        --registry-password {{ auth.get('password') }}
+    - failhard: True
+
+{{ macros.end_step('Login into registry') }}
+
+{% endif %}
+
+{{ macros.begin_step('Download ceph container image') }}
+download ceph container image:
+  cmd.run:
+    - name: |
+        cephadm --image {{ pillar['ceph-salt']['container']['images']['ceph'] }} pull
+    - failhard: True
+{{ macros.end_step('Download ceph container image') }}
+
+{{ macros.end_stage('Prepare to bootstrap the Ceph cluster') }}
+
 {{ macros.begin_stage('Bootstrap the Ceph cluster') }}
 
 {% set bootstrap_ceph_conf = pillar['ceph-salt'].get('bootstrap_ceph_conf', {}) %}
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephtools.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephtools.sls
@@ -2,7 +2,7 @@
 
 {% if 'cephadm' in grains['ceph-salt']['roles'] %}
 
-{{ macros.begin_stage('Prepare to bootstrap the Ceph cluster') }}
+{{ macros.begin_stage('Configure cephadm') }}
 
 {{ macros.begin_step('Install cephadm and other packages') }}
 
@@ -19,17 +19,6 @@ install cephadm:
 # in case any package instalation re-writes sudoers /etc/sudoers.d/<ssh_user>
 {{ macros.sudoers('configure sudoers after package instalation') }}
 
-{% if grains['id'] == pillar['ceph-salt'].get('bootstrap_minion') %}
-
-/var/log/ceph:
-  file.directory:
-    - user: ceph
-    - group: ceph
-    - mode: '0770'
-    - makedirs: True
-    - failhard: True
-{% endif %}
-
 {{ macros.end_step('Install cephadm and other packages') }}
 
 {{ macros.begin_step('Run "cephadm check-host"') }}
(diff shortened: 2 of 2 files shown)


sebastian-philipp (approved):

smithfarm:
I'm fine with this, though it seems like the "Prepare to bootstrap the Ceph cluster" stage would fit better in `ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls` now.

ricardoasmarques:
I've now remembered why we decided to pull images in ceph-salt before-hand... the idea was to guarantee that commands like `cephadm shell` don't pull the "wrong" image:

```
dpscalam15-master:~ # ceph config dump
WHO     MASK  LEVEL     OPTION                                 VALUE                                                                                        RO
global        basic     container_image                        registry.suse.de/suse/sle-15-sp2/update/products/ses7/milestones/containers/ses/7/ceph/ceph  * 
  mon         advanced  public_network                         172.16.0.0/23                                                                                * 
  mgr         advanced  mgr/cephadm/manage_etc_ceph_ceph_conf  true                                                                                         * 
  mgr         advanced  mgr/cephadm/migration_current          2                                                                                            * 
  mgr         advanced  mgr/orchestrator/orchestrator          cephadm 

dpscalam15-master:~ # cephadm shell
Trying to pull docker.io/ceph/ceph:v15...
...
```

In this example, `cephadm shell` pulled `docker.io/ceph/ceph:v15` instead of `registry.suse.de/suse/sle-15-sp2/update/products/ses7/milestones/containers/ses/7/ceph/ceph` because we don't have an image already pulled that `cephadm` can infer.

I'm afraid this prevents us from merging this PR.

ricardoasmarques:
Here is a PR to guarantee we will never have more than 10 minions pulling the ceph image at the same time: https://github.com/ceph/ceph-salt/pull/355

sebastian-philipp:
Would https://github.com/SUSE/ceph/pull/404 enable us to merge this?

ricardoasmarques:
> Would [SUSE/ceph#404](https://github.com/SUSE/ceph/pull/404) enable us to merge this?

If we merge that PR, should we consider removing the `/containes/images/ceph` config since it will be ignored?

```
master:~ # ceph-salt config /containers/images ls
o- images .................................................................................................................... [...]
  o- ceph ................................................ [registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph]
```

sebastian-philipp:
> > Would [SUSE/ceph#404](https://github.com/SUSE/ceph/pull/404) enable us to merge this?
> 
> If we merge that PR, should we consider removing the `/containes/images/ceph` config since it will be ignored?

Nothing is getting ignored. SUSE/ceph#404 simply provides a default, if 

* no image is passed to bootstrap, or
* there is no existing image to automatically deduce it.

ricardoasmarques:
> ...
> Nothing is getting ignored. [SUSE/ceph#404](https://github.com/SUSE/ceph/pull/404) simply provides a default, if
> 
> * no image is passed to bootstrap, or
> * there is no existing image to automatically deduce it.

Exactly, if ceph-salt dont pre-pull image on all hosts, when you ~~deploy a service~~ run `cephadm shell` on a host then the `default` image will be used, not the one specified in `ceph-salt config`. This is why I think this config will be useless if we merge this PR + https://github.com/SUSE/ceph/pull/404.

On the other hand, this image will be used in bootstrap and for service deployment.

Also, `cephadm shell` may use different images depending on the host we are running it (depends if the host already has services deployed or not).

smithfarm:
@sebastian-philipp I guess there is no harm in merging https://github.com/SUSE/ceph/pull/404 - we have to make our best effort in ceph-salt to ensure that all the cluster nodes get the "correct" image. What the "correct" image path is will vary from site to site, but I agree that in a supported SES7 cluster it will never start with "docker.io". So, let's merge https://github.com/SUSE/ceph/pull/404 by all means, but it doesn't have anything to do with the PR.

Now, regarding this PR: if we merged it, the bootstrap minion would pull the "correct" image path (defined as the one configured by the user), while the rest of the cluster nodes would get the hardcoded default image path, which might well be different. That wouldn't be good. Conclusion: I vote for closing this PR.

ricardoasmarques:
> ...
>..., while the rest of the cluster nodes would get the hardcoded default image path, which might well be different. ...

It depends:
1) If you deploy a service on the host and then run `cephadm shell`
    in this case, you will always pull the image defined in `ceph-salt config`

2) If you do it in the oposite order, run `cephadm shell` and then deploy a service
    in this case you will pull both, because `cephadm shell` will will pull the default hardcoded one, and then service deployment will pull the onde configured in `ceph-salt`

ricardoasmarques:
I agree that merging https://github.com/SUSE/ceph/pull/404 is a good idea anyway.

Note that even if `ceph-salt` pre-pull the image, it's sill possible that user will remove that image manually and then run `cephadm shell`... in this case, we don't want to pull the upstream default

ricardoasmarques:
Relates to https://github.com/ceph/ceph/pull/33781

We could resurrect `#33781`, but note that we are trying to avoid adding `cephadm` config file until really-really needed.

sebastian-philipp:
> Relates to [ceph/ceph#33781](https://github.com/ceph/ceph/pull/33781)
> 
> We could resurrect `#33781`, but note that we are trying to avoid adding `cephadm` config file until really-really needed.

cephadm is on its way to become a proper software. I.e. proper python module structure, proper packaging etc. At that point we should think about having a proper configuration file for cephadm. Having to do downstream patches is a bad idea in general. 

But that's pacific and not octopus.

sebastian-philipp:
> but https://github.com/SUSE/ceph/pull/404 doesn't have anything to do with the PR.

I'd disagree here. It's also about where we want to put our time and energy into. 

If we're ok with users maybe downloading the wrong downstream image and thus calling the wrong `ceph` binary when calling `cephadm shell`, then https://github.com/SUSE/ceph/pull/404 should suffice. 

If we we really cannot do that, then we might want to continue with  #355 . 

In general I'm aiming towards a future where we `cepahdm` should do the right thing in any case. Independent of ceph-salt.

ricardoasmarques:
If the following `cephadm` issue is fixed, then we can merge this PR: https://tracker.ceph.com/issues/47274

sebastian-philipp:
> If the following `cephadm` issue is fixed, then we can merge this PR: https://tracker.ceph.com/issues/47274

note that this would just fix a race: As soon as the crash daemon is deployed on that host, cephadm shell is *already* using the correct image.

ricardoasmarques:
> I'm fine with this, though it seems like the "Prepare to bootstrap the Ceph cluster" stage would fit better in `ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls` now.

Sure, done!

ricardoasmarques:
> > If the following `cephadm` issue is fixed, then we can merge this PR: https://tracker.ceph.com/issues/47274
> 
> note that this would just fix a race: As soon as the crash daemon is deployed on that host, cephadm shell is _already_ using the correct image.

This seems to be a good reason to merge this PR.

When a new host is added, `cephadm` will deploy `crash` daemon on it, and to do that, `cephadm` will pull the same image that was used during bootstrap.

For now, as soon as `crash` daemon is deployed, it's safe do run `cephadm shell`.

When https://tracker.ceph.com/issues/47274 is fixed, then it will be safe to run `cephadm shell` at any time.

PR #354, opened 2020-09-01
execute: implement test.ping sanity check

### LINKED ISSUE #331: ceph-salt status/apply: add Salt sanity check ("test.ping")
`ceph-salt status`/`ceph-salt apply` runs various sanity checks before applying the Salt Formula.

Ideally we would add a sanity check for every possible problem we can think of, because once the Salt Formula is running, it's difficult to return meaningful error messages to the user.

Here's an idea for a sanity check:

`ceph-salt status`/`ceph-salt apply` can enumerate all the minions by consulting the pillar (`ceph-salt config ls`). We know from experience that it's possible for one or more minions to be out of commission for various reasons:

1. minion node could be down
2. salt-minion.service might not be running on the minion
3. minion might not be reachable due to a networking issue
4. minion might be using a key that the master rejected - both salt-minion and salt-master have been known to cache keys, making it surprisingly difficult to recover from a rejected key situation

The `test.ping` sanity check would run `salt '*' test.ping` (or maybe use a grain for minion targeting?) in a loop, checking the minions that respond against the list of minions in the pillar. Once all are responding, it can continue. If it exhausts, say, 10 tries in a try-wait loop, it will fail and instruct the user to check that all minions are up and pingable from the master.

smithfarm:
Before this commit, if any minion is down or otherwise not responding to
the Salt Master, "ceph-salt apply" will fail with a bogus error message:

"Could not find ceph-salt formula. Please check if ceph-salt-formula
package is installed"

To address this, all configured minions are pinged and number of
responding minions is compared against the number of configured minions.
If any do not respond, an appropriate error message is displayed.

Fixes: https://github.com/ceph/ceph-salt/issues/331

### DIFF
--- a/ceph_salt/execute.py
+++ b/ceph_salt/execute.py
@@ -1366,6 +1366,31 @@ class CephSaltExecutor:
         return 0
 
     @staticmethod
+    def ping_minions():
+        # verify that all minions are alive
+        PP.println("Checking if all minions respond to ping...")
+        all_minions = PillarManager.get('ceph-salt:minions:all', [])
+        minion_count = len(all_minions)
+        PP.println("Pinging {} minions...".format(minion_count))
+        result = SaltClient.local_cmd('ceph-salt:member', 'test.ping', tgt_type='grain')
+        # result will be something like {'node3.ses7.test': True, 'master.ses7.test': True,
+        # 'node2.ses7.test': True, 'node1.ses7.test': True}
+        minions_responding = 0
+        retval = 0
+        for minion, response in result.items():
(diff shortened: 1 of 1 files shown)


smithfarm:
```
master:~ # ssh node3 systemctl stop salt-minion
master:~ # ceph-salt apply
Checking if all minions respond to ping...
Pinging 4 minions...
node3.ses7.test did not respond to ping
master:~ # echo $?
8
```

ricardoasmarques (approved):
LGTM

Issue #362, opened 2020-09-02
Use same SSH config that is used by cephadm


ricardoasmarques:
`ceph-salt` should use the same SSH config used by `cephadm`: https://github.com/ceph/ceph/blob/master/src/pybind/mgr/cephadm/module.py#L76-L82

Issue #360, opened 2020-09-02
Salt module for SSH executions


ricardoasmarques:
The following code is duplicated in a bunch of places:

```
"ssh -o StrictHostKeyChecking=no {}@{} "
```

We could implement a salt module to perform ssh executions, and have this code on a single place.

Issue #357, opened 2020-09-02
Always use SSH 'cephadm' user


ricardoasmarques:
SSH should not be configurable in `ceph-salt config`, instead, we should always use the 'cephadm' user.

PR #363, opened 2020-09-02
Always use cephadm SSH user (and more...)

ricardoasmarques:
This PR fixes the following issues:

- **Always use SSH 'cephadm' user** (https://github.com/ceph/ceph-salt/issues/357)
- **Add salt module for SSH executions** (https://github.com/ceph/ceph-salt/issues/360)
- **Set 'UserKnownHostsFile' and 'ConnectTimeout' on SSH connections** (https://github.com/ceph/ceph-salt/issues/362)
- **Rename '/etc/sudoers.d' file to avoid collision with cephadm**
- **Centralize SSH cleanup code**

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
+++ b/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
@@ -14,7 +14,16 @@ create ssh user:
       - users
     - failhard: True
 
-{{ macros.sudoers('configure sudoers') }}
+configure sudoers:
+  file.append:
+    - name: /etc/sudoers.d/ceph-salt
+    - text:
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph -s"
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph orch host add *"
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph orch host ok-to-stop *"
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph orch status --format=json"
+      - "cephadm ALL=NOPASSWD: /usr/bin/python3"
+      - "cephadm ALL=NOPASSWD: /usr/bin/rsync"
 
 # make sure .ssh is present with the right permissions
 create ssh dir:
--- a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
+++ b/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
@@ -2,45 +2,36 @@
 
 {{ macros.begin_stage('Distribute SSH keys') }}
 
-{% set ssh_user = pillar['ceph-salt']['ssh']['user'] %}
-{% set ssh_user_group = 'root' if ssh_user == 'root' else 'users' %}
-
-{% if ssh_user != 'root' %}
-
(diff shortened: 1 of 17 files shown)


smithfarm:
Does this depend on https://github.com/SUSE/ceph/pull/403 ?

ricardoasmarques:
> Does this depend on [SUSE/ceph#403](https://github.com/SUSE/ceph/pull/403) ?

Yes

ceph-salt-formula/salt/macros.yml
```diff
@@ -26,17 +26,14 @@
 {%- endmacro %}
 
 {% macro sudoers(name) -%}
-{%- set ssh_user = pillar['ceph-salt']['ssh']['user'] %}
-{% if ssh_user != 'root' %}
 {{name}}:
   file.append:
-    - name: /etc/sudoers.d/{{ssh_user}}
+    - name: /etc/sudoers.d/cephadm
```
smithfarm: You know, we could avoid the collision with the `/etc/sudoers.d/cephadm` installed by Ceph by simply using a different filename - e.g. `/etc/sudoers.d/ceph-salt`, or `/etc/sudoers.d/cephadm-ceph-salt`

smithfarm:
> > Does this depend on [SUSE/ceph#403](https://github.com/SUSE/ceph/pull/403) ?
> 
> Yes

Actually, not. I just realized that, in the filename `/etc/sudoer.d/cephadm`, the string "cephadm" is just a string. It carries no semantic meaning.

ceph-salt-formula/salt/macros.yml
ricardoasmarques: Good catch! Thanks!

ricardoasmarques:
This PR is now ready for review.

ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
```diff
@@ -2,63 +2,63 @@
...
-
-{{ macros.sudoers('configure sudoers') }}
+configure sudoers:
+  file.append:
+    - name: /etc/sudoers.d/ceph-salt
+    - text:
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph -s"
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph orch host add *"
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph orch host ok-to-stop *"
+      - "cephadm ALL=NOPASSWD: /usr/bin/ceph orch status --format=json"
```
smithfarm, line 24: instead of taxatively listing specific `ceph` CLI commands, would it make sense to replace all of these lines with:

```
"cephadm ALL=NOPASSWD: /usr/bin/ceph"
```

(Because, otherwise, we will have to edit this file whenever we include code in ceph-salt that runs a `ceph` command...)

smithfarm: I can't help thinking that users are going to want to log in as the `cephadm` user on the Salt Master node, and use it to administer the Ceph cluster (instead of being logged in as `root` to do that).

smithfarm (approved):

PR #367, opened 2020-09-04
Sanity-check time sync services when /time_server is disabled

### LINKED ISSUE #208: Disabling time server functionality causes "ceph-salt apply" to fail 100% of the time
It's possible to disable `time server` configuration in `ceph-salt`, but if we do it then we will hit the following error when applying the config:

Screenshot from 2020-04-29 16-03-59

smithfarm:
Users can set /time_server to "disabled", but if they do that,
they are responsible for ensuring that the time sync service is enabled
and running on all the minions before starting to deploy a Ceph cluster.

Now, you may ask: "cephadm check-host" already checks this, so why are
we duplicating that check here in ceph-salt?

The answer is: without a pre-flight check for this, a user might disable
/time_server and then run "ceph-salt apply", and in this case "ceph-salt
apply" will fail while trying to apply the Salt Formula, yet we are
trying to catch all possible failure conditions before actually applying
the Salt Formula...

Fixes: https://github.com/ceph/ceph-salt/issues/208

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -108,3 +108,52 @@ def probe_dns(*hostnames):
         if not ret_status:
             break
     return ret_status
+
+
+def probe_time_sync():
+    units = [
+        'chrony.service',  # 18.04 (at least)
+        'chronyd.service', # el / opensuse
+        'systemd-timesyncd.service',
+        'ntpd.service', # el7 (at least)
+        'ntp.service',  # 18.04 (at least)
+    ]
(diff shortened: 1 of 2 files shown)


smithfarm:
```
    master: /time_server is disabled. Will check if minions have a time_sync service enabled and running...
    master: Checking time sync service on 4 minions...
    master: Time sync issues detected on host(s) node3.ses7.test, master.ses7.test, node2.ses7.test, node1.ses7.test
    master: /time_server is disabled. In that case, a time sync service must be enabled and running on all minions. Please fix this issue and try again.
```

ricardoasmarques (approved):
lgtm

PR #382, opened 2020-09-09
Readme: update diagram

sebastian-philipp:
e.g. remove ceph-salt-formula

Signed-off-by: Sebastian Wagner <sebastian.wagner@suse.com>

See: https://github.com/ceph/ceph-salt/tree/29d261d85fd262b38c4d014ac44d7f77ad7d1d36#architecture


smithfarm:
This diagram says that ceph-salt is responsible for "Everything that isn't strictly Ceph related", but that isn't a true statement, is it?

I suggest to drop that line from the diagram.

sebastian-philipp:
sure about it? I mean it helps understanding what ceph-salt does

smithfarm:
Yes, pretty sure ceph-salt is not responsible for "everything not Ceph related". I mean, that's an infinite amount of stuff and what ceph-salt does is actually quite limited.

I don't think a vague statement helps to understand what ceph-salt does.

PR #383, opened 2020-09-09
Support bootstrap minion without admin role

### LINKED ISSUE #272: UX: Setting the bootstrap minion should also set the `admin` role and the `cephadm` role
When setting the bootstrap minion, this minion should automatically added to the `roles/admin` and `roles`cephadm` list.
Otherwise you have to do theses steps manually.

```
/ceph_cluster/roles/bootstrap> ls
o- bootstrap ........................................................................................................... [no minion]               
/ceph_cluster/roles/bootstrap> set node1.ses7-manual.com
Value set.
/ceph_cluster/roles/bootstrap> exit
master:~ # ceph-salt apply
Bootstrap minion must be 'Admin'
```

When then adding the same node to the admin role and trying to apply again, I get:
```
master:~ # ceph-salt apply
Minion 'node1.ses7-manual.com' has 'admin' role but not 'cephadm' role
```

So next manual step is:

```
/ceph_cluster/roles/cephadm> add node1.ses7-manual.com
Adding node1.ses7-manual.com...
1 minion added.
/ceph_cluster/roles/cephadm> exit

```

ricardoasmarques:
With this PR, the `bootstrap minion` will no longer be required to have the `admin` role.

The only role that is required for the `bootstrap minion` is the `cephadm` role because the `bootstrap minion` will belong to `ceph orch host ls`.

This PR will also add a check to guarantee that the configuration contains at least one admin minion.

Related PRs:
- https://github.com/SUSE/doc-ses/pull/529
- https://github.com/SUSE/sesdev/pull/461

Fixes: https://github.com/ceph/ceph-salt/issues/272

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -41,13 +41,14 @@ def ssh(host, cmd):
                                    "-i /home/cephadm/.ssh/id_rsa "
                                    "cephadm@{} \"{}\"".format(host, cmd))
 
-def sudo_rsync(src, dest):
+def sudo_rsync(src, dest, ignore_existing):
+    ignore_existing_option = '--ignore-existing ' if ignore_existing else ''
     return __salt__['cmd.run_all']("sudo rsync --rsync-path='sudo rsync' "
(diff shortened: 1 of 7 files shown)


ricardoasmarques:
cc @LenzGr

PR #397, opened 2020-09-17
Convert tags to repo_digest

### LINKED ISSUE #300: Unalias the ceph container image tag before sending it to cephadm
As https://jira.suse.com/browse/SES-2017 explains, it will not be allowed to specify a ":latest" tag (or any other tag that isn't absolute, final, immutable) on the ceph container image path. That also means it won't be allowed to specify an image path without an explicit tag (since that would default to ":latest").

Since this will be confusing and bothersome for customers who are not accustomed to entering a long, complicated version number (the container image tag will be something like this: `:15.2.4.89.3.43`, which may not seem complicated to us developers, but is difficult for many people to type in correctly, and having to type it in is certainly bothersome for anyone, developers included).

This feature request is for ceph-salt to be patched so it queries the registry to find out which full version number tag corresponds to the container image path given by the user. (Some tags, such as ":latest" are just pointers to immutable tags. The immutable tag will be visible once the container image is downloaded.) Once the immutable tag is known, ceph-salt can munge the container image path to use the immutable tag. By "immutable tag" we mean a tag that has six version number components x.y.z.a.b.c, e.g. `:15.2.4.89.3.43`.

ricardoasmarques:
**After https://github.com/ceph/ceph/pull/36432 ~~is backported to octopus~~**

---

This PR will set:
```
# ceph config set mgr mgr/cephadm/use_repo_digest true --force
```
So tags will be converted to repo_digest:

Screenshot from 2020-09-17 11-53-01

Fixes: https://github.com/ceph/ceph-salt/issues/300

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/apply/ceph-admin.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/ceph-admin.sls
@@ -16,6 +16,7 @@ configure cephadm mgr module:
         ceph cephadm registry-login -i /tmp/ceph-salt-registry-json
 {%- endif %}
         ceph config set mgr mgr/cephadm/manage_etc_ceph_ceph_conf true
+        ceph config set mgr mgr/cephadm/use_repo_digest true --force
     - failhard: True
 
 {{ macros.end_stage('Ensure cephadm MGR module is configured') }}


ricardoasmarques:
While https://github.com/ceph/ceph/pull/36432 is not backported, the new `mgr/cephadm/use_repo_digest` will be ignored, but everything will work as expected so we don't need to wait for the backport.

PR #407, opened 2020-09-23
Restart 'chronyd' service to apply chrony config

ricardoasmarques:
If 'chronyd' is already running before executing 'ceph-salt apply', then changes on '/etc/chrony.conf' are ignored, so chronyd must be restarted to ensure '/etc/chrony.conf' is loaded.


smithfarm (approved):

LenzGr (approved):
LGTM!

PR #415, opened 2020-10-02
Reduce ceph-salt pillar file permissions

ricardoasmarques:
BEFORE
```
master:~ # ls -l /srv/pillar/ceph-salt.sls
-rw-r--r-- 1 salt salt 180 Oct  2 14:05 /srv/pillar/ceph-salt.sls
```

AFTER
```
master:~ # ls -l /srv/pillar/ceph-salt.sls
-rw------- 1 salt salt 180 Oct  2 14:11 /srv/pillar/ceph-salt.sls
```

### DIFF
--- a/ceph-salt.spec
+++ b/ceph-salt.spec
@@ -145,9 +145,12 @@ Salt Formula to deploy Ceph clusters.
 %dir %attr(0755, root, salt) %{fdir}/states/
 %dir %attr(0755, root, salt) %{fdir}/metadata/
 %dir %attr(0755, root, root) %{_datadir}/%{fname}
+%dir %attr(0755, salt, salt) %{_datadir}/%{fname}/pillar
+%doc %attr(0644, salt, salt) %{_datadir}/%{fname}/pillar.conf.example
+%attr(0600, salt, salt) %{_datadir}/%{fname}/pillar/ceph-salt.sls
+%attr(0644, salt, salt) %{_datadir}/%{fname}/pillar/top.sls
 %{fdir}/states/
 %{fdir}/metadata/
-%{_datadir}/%{fname}
 
 
 %changelog
--- a/ceph_salt/salt_utils.py
+++ b/ceph_salt/salt_utils.py
@@ -263,6 +263,7 @@ class PillarManager:
                 file.write(content)
             file.write("\n")
         shutil.chown(full_path, "salt", "salt")
+        os.chmod(full_path, 0o600)
 
     @classmethod
     def _load(cls):


ceph-salt.spec
```diff
@@ -145,6 +145,7 @@ Salt Formula to deploy Ceph clusters.
 %dir %attr(0755, root, salt) %{fdir}/states/
 %dir %attr(0755, root, salt) %{fdir}/metadata/
 %dir %attr(0755, root, root) %{_datadir}/%{fname}
+%attr(0600, salt, salt) %{buildroot}%{_datadir}/%{fname}/pillar/ceph-salt.sls
 %{fdir}/states/
 %{fdir}/metadata/
 %{_datadir}/%{fname}
```
smithfarm: I don't think this spec file change will work. There are two ways to package files and directories under a particular directory: either list them all explicitly, or just list the directory. `ceph-salt-formula` packaging is using the latter technique. You are trying to mix them here.

ceph-salt.spec
```diff
@@ -145,6 +145,7 @@ Salt Formula to deploy Ceph clusters.
 %dir %attr(0755, root, salt) %{fdir}/states/
 %dir %attr(0755, root, salt) %{fdir}/metadata/
 %dir %attr(0755, root, root) %{_datadir}/%{fname}
+%attr(0600, salt, salt) %{buildroot}%{_datadir}/%{fname}/pillar/ceph-salt.sls
```
smithfarm: need to drop `%{buildroot}` here, and also see my comment below

ceph-salt.spec
smithfarm: Here are the full contents of this directory:

```
/usr/share/ceph-salt
/usr/share/ceph-salt/pillar
/usr/share/ceph-salt/pillar.conf.example
/usr/share/ceph-salt/pillar/ceph-salt.sls
/usr/share/ceph-salt/pillar/top.sls
```

So, instead of `%{_datadir}/%{fname}` (which just pulls it all in with the default ownership/permissions), we need to explicitly list all the files and directories. Something like this:

```
%dir %attr(0755, root, root) %{_datadir}/%{fname}
%dir %attr(0755, salt, salt) %{_datadir}/%{fname}/pillar
%doc %attr(0644, salt, salt) %{_datadir}/%{fname}/pillar/pillar.conf.example
%attr(0600, salt, salt) %{_datadir}/%{fname}/pillar/ceph-salt.sls
%attr(0644, salt, salt) %{_datadir}/%{fname}/pillar/top.sls
```

and drop the line `%{_datadir}/%{fname}` at the bottom of this `%files -n ceph-salt-formula` section.

smithfarm (approved):

PR #422, opened 2020-10-06
Do the right thing in FQDN environments

### LINKED ISSUE #417: hostnames cannot be fqdn
This isn't a feature request, but more for information and tracking.  Possibly the README should say that `hostnamectl fqdn` is not supported or add a validation.

Here's the behavior: I use vagrant for 14 VMs.  With other recent changes, I've had an issue where my hostnames were not resolving from dnsmasq and were assigned ip addresses as hostnames.  So, I used `hostnamectl fqdn` as part of the setup.  The first  failure is that `ceph-salt` calls cephadm which fails with `use --allow-fqdn-hostname` in the log file.

If you change the call in `cephbootstrap.sls`, then `ceph-salt apply` will complete; however, the installation will not have an admin keyring.  So the containers started and that's it.

I'm forcing short hostnames now which is working.

### LINKED ISSUE #420: ceph-salt cannot bootstrap a Ceph cluster in FQDN environments
### What do I mean by "FQDN environments"

In accordance with https://docs.ceph.com/en/latest/cephadm/concepts/#fully-qualified-domain-names-vs-bare-host-names 

a "FQDN environment" is an environment where:

* `hostname` returns the FQDN
* `hostname -s` returns the short ("bare") hostname

### Why do FQDN environments need to be supported

While there certainly are customer environments where all nodes are in the same domain and each node is therefore guaranteed to have a unique short ("bare") hostname, we cannot get away with supporting only these environments. Why not?

#### Reason 1

Imagine a hypothetical customer who has two hypothetical networks: "foo.org" and "baz.net" and the following conditions hold true:

* The networks are fully routed - packets pass freely from one to the other, with very low latency
* The customer wishes to deploy a Ceph cluster with 1 MON on "foo.org" and 2 MONs on "baz.net"
* The customer is using standard DNS.
* The nodes in "foo.org" are named "node1.foo.org, node2.foo.org, ..., noden.foo.org" and the nodes in "baz.net" are named "node1.baz.net, node2.baz.net, ..., noden.baz.net"

This is a valid use case and I believe SES6/DeepSea has no problem with it.

### Reason 2

Even a "standard" customer environment where all nodes are in the same FQDN with unique short hostnames might need to be migrated to a different network. Such a migration would of course need to be supported without any interruption of service. Nodes would have to be migrated one by one, while the cluster remains up and running, and the expectation would be that the cluster remains in HEALTH_OK status throughout the procedure.

During the procedure, the cluster will have some nodes in network "original.net" while others are in "new.net" - similar to the scenario described in "Reason 1", above.

### The Bug

When I run "cephadm bootstrap" in a FQDN environment, it complains that I should either change my environment so "hostname" returns the short hostname or provide the "--allow-fqdn-hostname" option to "cephadm bootstrap".

Now, changing my environment is not an option (see "Reason 1" and "Reason 2", above). So I re-run "cephadm bootstrap", providing the --allow-fqdn-hostname option as suggested. **But this fails because ceph-salt assumes the short hostname in several places. Also, ceph-salt does not currently detect FQDN environments so it has no way of telling how it should behave.**

### How I reproduced the bug

* https://github.com/SUSE/sesdev/pull/519 adds "--fqdn" option to "sesdev create"
* https://github.com/ceph/ceph-salt/pull/418 adds "--allow-fqdn-hostname" option to "cephadm bootstrap"
* https://github.com/ceph/ceph-salt/pull/419 changes `grains['host']` to `grains['fqdn']`

smithfarm:
If *all* minions have FQDN environment, then use FQDN.

If *no* minions have FQDN environment, then use short hostname.

Otherwise (mixed FQDN/non-FQDN or could not determine), do not apply the
Salt Formula.

Fixes: https://github.com/ceph/ceph-salt/issues/417
Fixes: https://github.com/ceph/ceph-salt/issues/420

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -176,3 +176,20 @@ def __check_unit(unit_name):
     elif out in ['failed', 'auto-restart']:
         state = 'error'
     return (enabled, installed, state)
+
+
+def hostname():
+    return socket.gethostname()
+
+
+def probe_fqdn():
+    """
+    Returns 'YES' if FQDN environment detected, 'NO' if not detected, or 'FAIL'
+    if detection was not possible.
+    """
+    retval = hostname()
+    if not retval:
+        return 'FAIL'
+    if '.' in retval:
+        return 'YES'
+    return 'NO'
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
@@ -58,6 +58,7 @@ download ceph container image:
 {% set bootstrap_ceph_conf = pillar['ceph-salt'].get('bootstrap_ceph_conf', {}) %}
 {% set bootstrap_ceph_conf_tmpfile = "/tmp/bootstrap-ceph.conf" %}
 {% set bootstrap_spec_yaml_tmpfile = "/tmp/bootstrap-spec.yaml" %}
+{% set my_hostname = salt['ceph_salt.hostname']() %}
 
 create static bootstrap yaml:
   cmd.run:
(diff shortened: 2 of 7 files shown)


ricardoasmarques:
Just to avoid confusion, can we drop the `CephNode.hostname` property from `core.py`? Looks like it's no being used and it may return a value different than the new `ceph_salt.hostname` module, so I believe it's better to remove it.

ceph-salt-formula/salt/_modules/ceph_salt.py
```diff
@@ -176,3 +177,26 @@ def __check_unit(unit_name):
     elif out in ['failed', 'auto-restart']:
         state = 'error'
     return (enabled, installed, state)
+
+
+def hostname():
+    try:
+        process = subprocess.run(['hostname'],
```
ricardoasmarques: Any reason to NOT use an implementation similar to the one from `cephadm`?

- https://github.com/ceph/ceph/blob/master/src/cephadm/cephadm#L2780-L2782
- https://github.com/ceph/ceph/blob/master/src/cephadm/cephadm#L1214-L1216

smithfarm: > Any reason to NOT use an implementation similar to the one from `cephadm`?

They appear to behave the same, so I guess not:

```
master:~ # hostname
master
master:~ # python3
Python 3.6.10 (default, Dec 19 2019, 15:48:40) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.gethostname()
'master'
>>> 
master:~ # hostnamectl --set-hostname master.ses7.test
hostnamectl: unrecognized option '--set-hostname'
master:~ # hostnamectl set-hostname master.ses7.test
master:~ # hostname
master.ses7.test
master:~ # python3
Python 3.6.10 (default, Dec 19 2019, 15:48:40) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.gethostname()
'master.ses7.test'
```

ricardoasmarques (approved):
lgtm

PR #432, opened 2020-10-13
Pillar targeting by minion id instead of grain

ricardoasmarques:
Overview
---
The `ceph-salt` pillar contains sensitive data (e.g., SSH private key), so we need to guarantee that only minions allowed to consume that information will be able to access it.

The problem
---
Until this PR, `ceph-salt` was using grain targetting to distribute the `ceph-salt` pillar:

```
# cat top.sls
base:	
  'ceph-salt:member':	
    - match: grain	
    - ceph-salt
```
Since grains can be set by the minions, nothing prevents a non-ceph-salt minion, registered to the same salt master, from setting the `ceph-salt:member grain` and access the `ceph-salt pillar` without "permission".

Example:
```
# cat /etc/salt/grains
cat: /etc/salt/grains: No such file or directory

# salt-call pillar.items
local:
    ----------


# printf "ceph-salt:\n  member: true" > /etc/salt/grains

# salt-call pillar.items
local:
    ----------
    ceph-salt:
        ----------
        bootstrap_ceph_conf:
            ----------
            global:
                ----------
                cluster_network:
                    192.168.200.0/24
< -- cut -->
```

The solution
---
The only grain that is considerer secure to use is the `id`, because changing that grain requires the master to re-accept the minion key, so we are now using minion `id` targetting to distribute `ceph-salt` pillar:

```
master:/srv/pillar # cat top.sls 
base:
{% include 'ceph-salt-top.sls' %}

master:/srv/pillar # cat ceph-salt-top.sls 
{% import_yaml "ceph-salt.sls" as ceph_salt %}
{% set ceph_salt_minions = ceph_salt.get('ceph-salt', {}).get('minions', {}).get('all', []) %}
{% if ceph_salt_minions %}
  {{ ceph_salt_minions|join(',') }}:
    - match: list
    - ceph-salt
{% endif %}
```
Which will guarantee that only minions listed in `ceph-salt:minions:all` pillar will be able to read `ceph-salt` pillar.

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/apply/init.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/init.sls
@@ -1,4 +1,4 @@
-{% if 'ceph-salt' in grains and grains['ceph-salt']['member'] %}
+{% if grains['id'] in pillar['ceph-salt']['minions']['all'] %}
 
 include:
     - ..reset
--- a/ceph-salt-formula/salt/ceph-salt/purge/init.sls
+++ b/ceph-salt-formula/salt/ceph-salt/purge/init.sls
@@ -1,4 +1,4 @@
-{% if 'ceph-salt' in grains and grains['ceph-salt']['member'] %}
+{% if grains['id'] in pillar['ceph-salt']['minions']['all'] %}
 
 check safety:
    ceph_salt.check_safety:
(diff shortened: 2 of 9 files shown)


ricardoasmarques:
~~Package for testing: https://build.opensuse.org/package/show/home:rimarques:branches:filesystems:ceph:octopus/ceph-salt~~

PR #433, opened 2020-10-13
Do not create ceph-salt pillar file during installation

ricardoasmarques:
Files that will be edited by the user should not be created during package installation, otherwise, they will be overridden during package update:

```
master:~ # rpm -q ceph-salt-formula
ceph-salt-formula-15.2.13+1602621538.ga6a6c18-lp152.3.1.noarch

master:~ # cat /usr/share/ceph-salt/pillar/ceph-salt.sls
ceph-salt:
  test: 123

master:~ # zypper up ceph-salt-formula
...

master:~ # rpm -q ceph-salt-formula
ceph-salt-formula-15.2.13+1602623476.ga6a6c18-lp152.4.1.noarch

master:~ # cat /usr/share/ceph-salt/pillar/ceph-salt.sls
ceph-salt:

master:~ #
```

### DIFF
--- a/ceph-salt.spec
+++ b/ceph-salt.spec
@@ -94,38 +94,6 @@ mkdir -p %{buildroot}%{fdir}/metadata/%{fname}/
 cp -R ceph-salt-formula/salt/* %{buildroot}%{fdir}/states/
 cp ceph-salt-formula/metadata/* %{buildroot}%{fdir}/metadata/%{fname}/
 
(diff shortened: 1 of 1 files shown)


ricardoasmarques:
Also, note that `ceph-salt config` will write the `ceph-salt.sls` pillar file on the first `pillar_roots` location, which is `/srv/pillar` by default, so creating a `/usr/share/ceph-salt/pillar/ceph-salt.sls` file that will not be used will just cause confusion.

PR #437, opened 2020-10-15
apply: Ensure mgr/cephadm/container_init is set to true

smithfarm:
By default, the "mgr/cephadm" orchestrator does not provide the "--init"
option when starting containers. It can by made to do this, though, by
explicitly setting an option ("mgr/cephadm/container_init").

On freshly deployed SES7 clusters, this option will have already been
set by "cephadm bootstrap --container-init", in which case this line
will be a noop. But it is still needed for upgraded clusters.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1177588


smithfarm:
I guess `cephadm adopt` could take care of setting this option, just like `cephadm bootstrap` does. But having it here in ceph-salt can't hurt...

ricardoasmarques (approved):
lgtm

PR #438, opened 2020-10-15
Add ceph-osd sysctl settings

mgfritch:
fs.aio-max-nr = 1048576
kernel.pid_max = 4194304

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/apply/sysctl.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/sysctl.sls
@@ -1,0 +1,28 @@
+{% import 'macros.yml' as macros %}
+
+{% if 'cephadm' in grains['ceph-salt']['roles'] %}
+
+{{ macros.begin_stage('Configure sysctl') }}
+
+/usr/lib/sysctl.d/90-ceph-salt.conf:
+  file.managed:
+    - source:
+        - salt://ceph-salt/apply/files/90-ceph-salt.conf.j2
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: '0644'
+    - makedirs: True
+    - backup: minion
+    - failhard: True
+
+reload sysctl:
+  cmd.run:
+    - name: "sysctl --system"
+
+{{ macros.end_stage('Configure sysctl') }}
+
+{% endif %}
+
+sysctl:
+  test.nop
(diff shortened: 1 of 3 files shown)


ricardoasmarques:
Note that every time `ceph-salt apply` is executed, the `/etc/sysctl.d/90-ceph.conf` file will be overwritten which means that if these default values are not ideal for every environment, we will need to make those values configurable via, `ceph-salt config` (because, since it's managed by `ceph-salt`, the user should not edit `/etc/sysctl.d/90-ceph.conf` file manually).

But I think we can make those values configurable in a future PR, if needed? Agree?

smithfarm:
@ricardoasmarques Upstream doesn't make this configurable, either - the file is created by installing an RPM (which no longer works because the RPM ends up in a container, hence this fix). According to `man sysctl.d`:

    Packages should install their configuration files in /usr/lib/. Files in /etc/ are reserved for the 
    local administrator, who may use this logic to override the configuration files installed by
    vendor packages.

smithfarm (approved):

ceph-salt-formula/salt/ceph-salt/apply/sysctl.sls
```diff
@@ -0,0 +1,28 @@
+{% import 'macros.yml' as macros %}
+
+{% if 'cephadm' in grains['ceph-salt']['roles'] %}
+
+{{ macros.begin_stage('Configure sysctl') }}
+
+/etc/sysctl.d/90-ceph.conf:
```
smithfarm: Oops. `/etc/sysctl.d` is reserved for local administrators, so they can override vendor settings like the ones we are creating in this PR.

I also think "ceph-salt" is a better label than "ceph", since the name `ceph-salt` will more effectively signal which tool is responsible for creating the file.

So, could you please change this to `/usr/lib/sysctl.d/90-ceph-salt.conf`?

smithfarm (changes requested):

ceph-salt-formula/salt/ceph-salt/apply/sysctl.sls
mgfritch: eh debatable, salt is a configuration management/remote administration tool and not a package manager (RPM).

mgfritch:
> Note that every time `ceph-salt apply` is executed, the `/etc/sysctl.d/90-ceph.conf` file will be overwritten which means that if these default values are not ideal for every environment, we will need to make those values configurable via, `ceph-salt config` (because, since it's managed by `ceph-salt`, the user should not edit `/etc/sysctl.d/90-ceph.conf` file manually).
> 
> But I think we can make those values configurable in a future PR, if needed? Agree?

yes I thought about adding values that come from pillar, but we aren't doing that for the latency and throughput settings either .. adding via a future PR seems aok since these can always be overridden manually via `/etc/sysctl.conf` ...

smithfarm (approved):

PR #439, opened 2020-10-19
Support multiple time servers

### LINKED ISSUE #426: ceph-salt lets user configure only one internal time source. This then becomes a single source of failure.
The ceph-salt documentation says:

    If your site already has a single source of time, specify the host name of the time source:

    ceph-salt config /time_server/server_hostname set time-server.example.com

But the SES6 Admin Guide says:

"We recommend synchronizing all Ceph cluster nodes with **at least three** reliable time sources that are located on the internal network. The internal time sources can point to a public time server or have their own time source."

(emphasis added)

As currently implemented, ceph-salt pushes users toward having only **one** time source on the internal network. This becomes a single source of failure.

I think we need to redesign it to make it possible for the user to give multiple internal time sources.

ricardoasmarques:
With this PR, it will be possible to specify multiple time servers:

```
master:~ # ceph-salt config /time_server ls
o- time_server ....................................................... [enabled]
  o- external_servers ...................................................... [1]
  | o- pool.ntp.org ...................................................... [...]
  o- servers ............................................................... [2]
  | o- node1.ses7.test ................................................... [...]
  | o- node2.ses7.test ................................................... [...]
  o- subnet ................................................... [10.20.147.0/24]
```

Note that this PR changes pillar structure, but `ceph-salt` does not support automatic pillar structure updates yet, which means that if you are upgrading from a previous `ceph-salt` version you need to re-configure timeserver or do the following replace in `/srv/pillar/ceph-salt.sls` manually:

**Before:**
```
server_hosts: <host>
```
**After:**
```
server_hosts:
- <host>
```

The config node was renamed too:
**Before:**
```
> /time_server/server_hostname
```
**After:**
```
> /time_server/servers
```

Fixes: https://github.com/ceph/ceph-salt/issues/426

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ sync its clock against one or more external servers (e.g. "pool.ntp.org"), and
 all other nodes in the cluster will sync their clocks against this node:
 
 ```
-/time_server/server_hostname set <minion id of the admin node>
+/time_server/servers add <minion id of the admin node>
 /time_server/external_servers add pool.ntp.org
 ```
 
@@ -118,7 +118,7 @@ can be useful when setting up a cluster at a site that already has a single
 source of time:
 
 ```
-/time_server/server_hostname set <hostname>
+/time_server/servers add <hostname>
 ```
 
 (In this case, the on-site time server is assumed to be already configured and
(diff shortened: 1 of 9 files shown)


ricardoasmarques:
README must be updated

ricardoasmarques:
> README must be updated

Done

PR #441, opened 2020-10-23
Drop the `container-init` flag during bootstrap

mgfritch:
cephadm will now start containers using an init process by default

After: https://github.com/ceph/ceph/pull/37764


ricardoasmarques:
If https://github.com/ceph/ceph/pull/37764#issuecomment-715254403 is implemented, we don't have to change ceph-salt, which makes it easier to submit Maintenance Updates because we don't have to sync `ceph` and `ceph-salt` package submissions.

mgfritch:
I think we will eventually run into the issue of sync'ing `ceph` and `ceph-salt` anyway?

mgfritch:
ceph/ceph#37764 adds backward compat for the legacy `--container-init` flag

PR #445, opened 2021-01-25
ceph-salt-formula: use pkg.installed in apply only

smithfarm:
It's not a good idea to put the pkg.installed state in sls files that will be
run often. This is because pkg.installed triggers a "zypper refresh" every time
it is run, and "zypper refresh" (a) takes a long time to run and (b) is prone to
failure.

Since we know that "ceph-salt apply" must completely successfully for ceph-salt
to be usable, we can move all pkg.installed invocations there. Elsewhere,
a simple file.exists check is enough to assert that the relevant RPM package is
still installed on the minion.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1177171

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
+++ b/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
@@ -14,10 +14,9 @@ create ssh user:
       - users
     - failhard: True
 
-install sudo:
-  pkg.installed:
-    - pkgs:
-      - sudo
+sudo sanity:
+  file.exists:
+    - name: /etc/sudoers.d
     - failhard: True
 
 configure sudoers:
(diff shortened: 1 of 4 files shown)


ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
```diff
@@ -14,10 +14,9 @@ create ssh user:
       - users
     - failhard: True
 
-install sudo:
-  pkg.installed:
-    - pkgs:
-      - sudo
+sudo sanity:
```
ricardoasmarques, line 17: On `ceph-salt apply`, this will be executed before `sudo` installation on `software.sls`, which may cause `ceph-salt apply` to always fail (until `sudo` is manually installed), see:

https://github.com/ceph/ceph-salt/blob/master/ceph-salt-formula/salt/ceph-salt/apply/init.sls#L5-L10

smithfarm: That sucks. Is there a way to check for the existence of this file in the Jinja rendering stage? Then we could run pkg.installed only if it's actually needed.

PR #455, opened 2021-06-22
README.md: fix broken cephadm link

### LINKED ISSUE #454: Broken link in README.md for cephadm
The link for cephadm at the top of the doc points to https://docs.ceph.com/docs/master/mgr/cephadm/ This is a broken link.

Is https://docs.ceph.com/en/latest/cephadm/#cephadm a better choice?

smithfarm:
Fixes: https://github.com/ceph/ceph-salt/issues/454

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,4 +155,4 @@ man ceph-salt
 
 [ceph]: https://ceph.io/
 [salt]: https://www.saltstack.com/
-[cephadm]: https://docs.ceph.com/docs/master/mgr/cephadm/
+[cephadm]: https://docs.ceph.com/en/octopus/cephadm/


README.md
```diff
@@ -155,4 +155,4 @@ man ceph-salt
 
 [ceph]: https://ceph.io/
 [salt]: https://www.saltstack.com/
-[cephadm]: https://docs.ceph.com/docs/master/mgr/cephadm/
+[cephadm]: https://docs.ceph.com/en/octopus/cephadm/
```
sebastian-philipp, line 158: The idea is to change readthedocs to make https://docs.ceph.com/en/latest/cephadm/ point to the latest *stable* release. What do you think of pointing to https://docs.ceph.com/en/latest/cephadm/ ?. This would fix the 404

smithfarm: @sebastian-philipp If ceph-salt were guaranteed to always be tested on, and known to work with, the latest cephadm, then I would have no objections.

I have used ceph-salt only with octopus cephadm. That's why I wrote "octopus" instead of "latest", here.

smithfarm: > This would fix the 404

My patch also fixes the 404, no?

smithfarm: My concern is that the latest cephadm might differ semantically from the cephadm ceph-salt is currently supporting. Or if it doesn't differ semantically now, it might begin to differ at some point in the future, as the cephadm developers add features and refactor the code base.

Once ceph-salt is confirmed to support a more recent cephadm (e.g. pacific), we can change this link to point to it.

But to have it just point to whatever is the latest cephadm is misleading IMO.

PR #456, opened 2021-08-02
ceph-salt-formula/cephbootstrap: drop --container-init

smithfarm:
As of 15.2.13, --container-init no longer works and --init is to be used
instead.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1188979

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
@@ -106,7 +106,6 @@ run cephadm bootstrap:
                 --allow-fqdn-hostname \
                 --apply-spec {{ bootstrap_spec_yaml_tmpfile }} \
                 --config {{ bootstrap_ceph_conf_tmpfile }} \
-                --container-init \
 {%- if not pillar['ceph-salt']['dashboard']['password_update_required'] %}
                 --dashboard-password-noupdate \
 {%- endif %}


smithfarm:
Testing with `sesdev create ses7 --single-node --non-interactive --qa-test --ceph-salt-repo https://github.com/smithfarm/ceph-salt --ceph-salt-branch wip-init`

sebastian-philipp:
you're sure about it? 

https://github.com/ceph/ceph/blob/a75ca1736d967fa82e006ed719cd738de7f4fbc5/src/cephadm/cephadm#L5750

and 

https://github.com/ceph/ceph/blob/5a784c87b94e0ea996e88d93f2de2a1a9497d277/src/cephadm/cephadm#L7952

sebastian-philipp:
better now :smile: You want to update the PR title?

sebastian-philipp (approved):

smithfarm:
@sebastian-philipp No, I'm not sure. This is not my cup of tea. All I know is: as of 15.2.13, we are no longer getting an init process in the containers.

sebastian-philipp:
> @sebastian-philipp No, I'm not sure. This is not my cup of tea. All I know is: as of 15.2.13, we are no longer getting an init process in the containers.

odd. Can you verify that the `unit.run` file contains the `--init` argument for podman?

smithfarm:
@sebastian-philipp With podman 2.2.1:

```
smithfarm@wilbur:~> podman run -it --init 733178a70f32 /bin/bash
d344a99d2a30:/ # ps ax
  PID TTY      STAT   TIME COMMAND
    1 pts/0    Ss     0:00 /dev/init -- /bin/bash
    7 pts/0    S      0:00 /bin/bash
   29 pts/0    R+     0:00 ps ax
```

With podman 2.1.1:

```
master:~ # podman run -it --init 20db24e925ab /bin/bash
Error: error configuring network namespace for container d6b38f9f2c1467058c89d6a30cb0cdb6f1ad994a6c96a759cff74771ab3bbefd: No CNI configuration file in /etc/cni/net.d/. Has your network provider started?
```

I guess we need podman 2.2.0 or greater for `--init` to work?

smithfarm:
@sebastian-philipp Thanks for looking. It seems we will need podman 2.2.x for `--init` ... correct?

sebastian-philipp:
should bet there. We verified `podman --init` back in September. podman did support it back then already.

smithfarm:
> > @sebastian-philipp No, I'm not sure. This is not my cup of tea. All I know is: as of 15.2.13, we are no longer getting an init process in the containers.
> 
> odd. Can you verify that the `unit.run` file contains the `--init` argument for podman?

I'm afraid I have no idea where in the filesystem to look for that file?

sebastian-philipp:
The file mentioned here: https://docs.ceph.com/en/latest/cephadm/troubleshooting/#manually-running-containers

smithfarm:
> Can you verify that the `unit.run` file contains the `--init` argument for podman?

I can verify that it is completely missing:

```
master:/var/lib/ceph/6a0ee6a2-f3a4-11eb-9c62-525400feefac # grep init osd.0/unit.run
master:/var/lib/ceph/6a0ee6a2-f3a4-11eb-9c62-525400feefac # grep init mon.master/unit.run
master:/var/lib/ceph/6a0ee6a2-f3a4-11eb-9c62-525400feefac # 
```

That explains the behavior (thanks!), but I don't understand why cephadm 15.2.13 is not putting `--init` into the unit.run files?

smithfarm:
@sebastian-philipp Maybe octopus is missing some backport?

smithfarm:
@sebastian-philipp Full disclosure: it was my understanding that, as of Octopus 15.2.13, cephadm should be adding `--init` automatically, so it should no longer be necessary to send the `--container-init` option to `cephadm bootstrap`. Also, it should no longer be necessary to carry the three downstream patches that were needed to make `--container-init` actually work pre-15.2.13.

So I dropped those patches. Later, I found that `--init` is no longer passed to podman, even though ceph-salt is passing the `--container-init` option to `cephadm bootstrap` and the container image has ceph 15.2.13.

So there is definitely a bug here somewhere. I just don't know where :-(

sebastian-philipp:
I think the ceph-salt code as it is *should* work ok. Do you know why the init process is missing on the customer? is the --init flag missing or is there a different reason?

smithfarm:
The --init flag is missing in the unit.run files because we dropped the three downstream patches that were making it appear there. In 15.2.13, the --init is supposed to appear in unit.run automatically, but it doesn't.

I think there must be one or more patches missing in upstream octopus.

ml8mr:
> The --init flag is missing in the unit.run files because we dropped the three downstream patches that were making it appear there. In 15.2.13, the --init is supposed to appear in unit.run automatically, but it doesn't.
> 
> I think there must be one or more patches missing in upstream octopus.

I wonder if the problem is due to [this commit](https://github.com/SUSE/ceph/commit/82d30afc413023a6cbb19f6529cae3c1e7e63a1d), which sets container_init to False due to Octopus supporting podman versions less than 2.0.

If I manually set CONTAINER_INIT=True, I see podman being executed with --init.

ml8mr:
> @sebastian-philipp Thanks for looking. It seems we will need podman 2.2.x for `--init` ... correct?

My SLE15 SP2 environment is running podman 2.1.1 and it supports --init.

sebastian-philipp:
> > The --init flag is missing in the unit.run files because we dropped the three downstream patches that were making it appear there. In 15.2.13, the --init is supposed to appear in unit.run automatically, but it doesn't.
> > I think there must be one or more patches missing in upstream octopus.
> 
> I wonder if the problem is due to [this commit](https://github.com/SUSE/ceph/commit/82d30afc413023a6cbb19f6529cae3c1e7e63a1d), which sets container_init to False due to Octopus supporting podman versions less than 2.0.
> 
> If I manually set CONTAINER_INIT=True, I see podman being executed with --init.

Which means the error is somewhere either in ceph-salt or in cephadm. This is great news. I wonder why things broke. According to 

https://github.com/ceph/ceph/blob/a75ca1736d967fa82e006ed719cd738de7f4fbc5/src/cephadm/cephadm#L6039-L6043

we should just need to add `--container-init` to bootstrap in order to make it work. 

> > @sebastian-philipp Thanks for looking. It seems we will need podman 2.2.x for `--init` ... correct?
> 
> My SLE15 SP2 environment is running podman 2.1.1 and it supports --init.

Right. Octopus also supports podman 1, which unfortunately does not, if I remember correctly.

smithfarm:
OK, so if I understand correctly, in an Octopus environment that is guaranteed to have podman >= 2, what is needed is:

1. revert 82d30afc413023a6cbb19f6529cae3c1e7e63a1d
2. close this PR
3. provide the `--container-init` option to `cephadm bootstrap`

@sebastian-philipp Is that also your understanding?

sebastian-philipp:
> OK, so if I understand correctly, what is needed is:
> 
> 1. revert 82d30afc413023a6cbb19f6529cae3c1e7e63a1d

I don't think you need to do that. `bootstrap --contianer-init` overwrites this default. I have the suspicion that this broke in octopus somehow. 

> 2. close this PR
> 3. continue providing the `--container-init` option to `cephadm bootstrap`

yeah 

> 
> @sebastian-philipp Is that also your understanding?

smithfarm:
@sebastian-philipp Ah... So I just need to wait until octopus gets fixed.

Would it make sense to add a regression test for this? Given that the test only needs to bootstrap with `--container-init` and look for the presence of `--init` in the unit.run files, it should be feasible to implement it.

sebastian-philipp:
> @sebastian-philipp Ah... So I just need to wait until octopus gets fixed.

Could you verify that this is in fact broken in octopus? And if it is broken in octopus, could you call `ceph config dump`?

smithfarm:
> Could you verify that this is in fact broken in octopus? And if it is broken in octopus, could you call `ceph config dump`?

I don't have any easy way to deploy upstream octopus. But I'll think about how I could do that.

For now, I confirmed what @ml8mr reported, i.e. that reverting 82d30afc413023a6cbb19f6529cae3c1e7e63a1d is all that is needed to fix the bug in an environment that is guaranteed to have podman 2. So this PR is currently not needed.

smithfarm:
> reverting 82d30afc413023a6cbb19f6529cae3c1e7e63a1d is all that is needed to fix the bug in
> an environment that is guaranteed to have podman 2.

The upstream octopus fix will probably not be that simple, since upstream octopus apparently still has to deal with the possibility that the environment might have an ancient, pre-2.0 podman.

PR #458, opened 2021-09-02
.github/workflows: add tox and more python versions

kshtsk:
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.com>

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -19,10 +19,10 @@ jobs:
         pip install -r requirements.txt
     - name: Lint with pylint
       run: |
-        pylint sesboot
+        pylint ceph_salt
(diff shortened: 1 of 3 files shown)


kshtsk:
@p-se I think we can remove "Travis CI - Pull Request" from requirements.

p-se (approved):
I wonder if we need testing.yml to directly call pytest as this should be done by tox, too (in tox.yml). But as those tests are 
1. working (as compared to the checks from travis) and
2. do not consume a significant amount of resources,

it should be safe to merge them (both) and deactivate Travis in favor of these tests.

sebastian-philipp:
yeah, let's move away form Travis for now

PR #461, opened 2021-09-27
Rely on cephadm package for cephadm user creation

### LINKED ISSUE #460: cephadm user/group created by ceph-salt inconsistent with user created by cephadm package
If you installed the RPM package of `cephadm`, it will create a cephadm user and group automatically if such don't already exist:

https://github.com/ceph/ceph/blob/d4fc4ecf0aac78f1c767e7a77599364bdf29e062/ceph.spec.in#L1548-L1551

```sh
getent group cephadm >/dev/null || groupadd -r cephadm
getent passwd cephadm >/dev/null || useradd -r -g cephadm -s /bin/bash -c "cephadm user for mgr/cephadm" -d %{_sharedstatedir}/cephadm cephadm
```

Note that the user and group are both marked system (`-r`), the default gid of the cephadm user is the cephadm group, and the home directory is /var/lib/cephadm.

Compare with ceph-salt:

https://github.com/ceph/ceph-salt/blob/e646be770c40c348c2a10ffaa23149a0ad81811a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls#L5-L15

This:
* creates the "users" group (which almost certainly already exists)
* puts the cephadm user in that group
* neither user nor group are marked system
* the home directory is /home/cephadm

Ideally, we would change ceph-salt to match the cephadm package, as follows:

```diff --git a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls b/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
index a796fcc..090c67d 100644
--- a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
+++ b/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
@@ -4,14 +4,15 @@
 
 create ssh user group:
   group.present:
-    - name: users
+    - name: cephadm
+    - system: True
 
 create ssh user:
   user.present:
     - name: cephadm
-    - home: /home/cephadm
-    - groups:
-      - users
+    - gid: cephadm
+    - system: True
+    - home: /var/lib/cephadm
     - failhard: True
 
 install sudo:
@@ -36,7 +37,7 @@ create ssh dir:
   file.directory:
     - name: /home/cephadm/.ssh
     - user: cephadm
-    - group: users
+    - group: cephadm
     - mode: '0700'
     - makedirs: True
     - failhard: True
@@ -46,7 +47,7 @@ create ceph-salt-ssh-id_rsa:
   file.managed:
     - name: /home/cephadm/.ssh/id_rsa
     - user: cephadm
-    - group: users
+    - group: cephadm
     - mode: '0600'
     - contents_pillar: ceph-salt:ssh:private_key
     - failhard: True
@@ -56,7 +57,7 @@ create ceph-salt-ssh-id_rsa.pub:
   file.managed:
     - name: /home/cephadm/.ssh/id_rsa.pub
     - user: cephadm
-    - group: users
+    - group: cephadm
     - mode: '0644'
     - contents_pillar: ceph-salt:ssh:public_key
     - failhard: True
```

Unfortunately, we can't do that, because it won't work for existing deployments, which already have a cephadm user. We can't change the home directory (because that would require moving files), and we can't change the default gid (because that would require changing file ownership in the home directory). Well, we could implement salt states to _try_ to do those things, but alas, on a running system, the home directory is in use, so this is all just going to be horrible.

I think the closest we can get to ideal is this:

```diff --git a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls b/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
index a796fcc..eac37dc 100644
--- a/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
+++ b/ceph-salt-formula/salt/ceph-salt/common/sshkey.sls
@@ -4,14 +4,16 @@
 
 create ssh user group:
   group.present:
-    - name: users
+    - name: cephadm
+    - system: True
 
 create ssh user:
   user.present:
     - name: cephadm
+    - system: True
     - home: /home/cephadm
     - groups:
-      - users
+      - cephadm
     - failhard: True
 
 install sudo:
@@ -36,7 +38,7 @@ create ssh dir:
   file.directory:
     - name: /home/cephadm/.ssh
     - user: cephadm
-    - group: users
+    - group: cephadm
     - mode: '0700'
     - makedirs: True
     - failhard: True
@@ -46,7 +48,7 @@ create ceph-salt-ssh-id_rsa:
   file.managed:
     - name: /home/cephadm/.ssh/id_rsa
     - user: cephadm
-    - group: users
+    - group: cephadm
     - mode: '0600'
     - contents_pillar: ceph-salt:ssh:private_key
     - failhard: True
@@ -56,7 +58,7 @@ create ceph-salt-ssh-id_rsa.pub:
   file.managed:
     - name: /home/cephadm/.ssh/id_rsa.pub
     - user: cephadm
-    - group: users
+    - group: cephadm
     - mode: '0644'
     - contents_pillar: ceph-salt:ssh:public_key
     - failhard: True
```

That'll give a cephadm user which actually is still in the default "users" group (because that's what the underlying `useradd` command will do), but is _also_ in the cephadm group. And, at least the cephadm group is created correctly.

Why am I proposing this change? Because in locked-down environments, where `/etc/ssh/sshd_config` specifies a limited set of groups in `AllowGroups`, we need the cephadm user to be in the cephadm group, so _that_ group can be added to AllowGroups, rather than adding "users" and opening things up too much.

tserong:
Rather than having ceph-salt create the cephadm user and group (with
home directory and gid inconsistent with the user the cephadm package
itself wants to create), we're now installing the cephadm package on
all minions, and relying on *it* to create the user and group.
ceph-salt will then pick up the cephadm user's home directory
automatically.

For new deployments, this will result in the cephadm user created as
a system user with gid=cephadm home=/var/lib/ceph as expected.  For
existing deployments upgraded from previous versions of ceph-salt,
everything continues to work with the old, existing, cephadm user
(home=/home/cephadm, gid=users), because we're now just picking up
the user's attributes at runtime, rather than hard-coding them.

Fixes: https://github.com/ceph/ceph-salt/issues/460

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -38,12 +38,13 @@ def ssh(host, cmd, attempts=1):
     assert attempts > 0
     attempts_count = 0
     retry = True
+    cephadm_home = __salt__['user.info']('cephadm')['home']
     while retry:
         ret = __salt__['cmd.run_all']("ssh -o StrictHostKeyChecking=no "
                                       "-o UserKnownHostsFile=/dev/null "
                                       "-o ConnectTimeout=30 "
-                                      "-i /home/cephadm/.ssh/id_rsa "
-                                      "cephadm@{} \"{}\"".format(host, cmd))
+                                      "-i {}/.ssh/id_rsa "
+                                      "cephadm@{} \"{}\"".format(cephadm_home, host, cmd))
         attempts_count += 1
         retcode = ret['retcode']
         # 255: Connection closed by remote host
@@ -60,12 +61,13 @@ def ssh(host, cmd, attempts=1):
 
 def sudo_rsync(src, dest, ignore_existing):
     ignore_existing_option = '--ignore-existing ' if ignore_existing else ''
+    cephadm_home = __salt__['user.info']('cephadm')['home']
     return __salt__['cmd.run_all']("sudo rsync --rsync-path='sudo rsync' "
                                    "-e 'ssh -o StrictHostKeyChecking=no "
                                    "-o UserKnownHostsFile=/dev/null "
                                    "-o ConnectTimeout=30 "
-                                   "-i /home/cephadm/.ssh/id_rsa' "
-                                   "{}{} {} ".format(ignore_existing_option, src, dest))
+                                   "-i {}/.ssh/id_rsa' "
+                                   "{}{} {} ".format(cephadm_home, ignore_existing_option, src, dest))
 
 
 def get_remote_grain(host, grain):
--- a/ceph-salt-formula/salt/ceph-salt/apply/ceph-admin.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/ceph-admin.sls
@@ -9,8 +9,8 @@
 configure cephadm mgr module:
   cmd.run:
     - name: |
-        ceph cephadm set-priv-key -i /home/cephadm/.ssh/id_rsa
-        ceph cephadm set-pub-key -i /home/cephadm/.ssh/id_rsa.pub
+        ceph cephadm set-priv-key -i {{ salt['user.info']('cephadm').home }}/.ssh/id_rsa
+        ceph cephadm set-pub-key -i {{ salt['user.info']('cephadm').home }}/.ssh/id_rsa.pub
         ceph cephadm set-user cephadm
 {%- if auth %}
         ceph cephadm registry-login -i /tmp/ceph-salt-registry-json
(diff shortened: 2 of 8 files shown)


sebastian-philipp (approved):

tserong:
The various tox failures are unrelated to this PR. Cryptodome is failing to load a shared library ("OSError: Cannot load native module 'Cryptodome.Cipher._raw_des': Not found '_raw_des.cpython-38-x86_64-linux-gnu.so', Not found '_raw_des.abi3.so', Not found '_raw_des.so'"), because https://github.com/Legrandin/pycryptodome/commit/47bca2b38665696d932df487f49a17b240f9c886 uses `os.path.isfile()` to check if the shared library exists, and AFAICT this fails in the test environment because of the presence of pyfakefs. I'm not yet sure how to fix/workaround this.

PR #462, opened 2021-09-28
tests: add source dir as real directory to fake fs

tserong:
We use Cryptodome for generating SSH keys, and that in turn tries to
dynamically load a shared library to do the work.  Recently (2021-09-19)
the load_pycryptodome_raw_lib() function was changed to use os.path.isfile()
to check if the shared library exists before loading it:

https://github.com/Legrandin/pycryptodome/commit/47bca2b38665696d932df487f49a17b240f9c886

The os.path.isfile() check fails when used inside a pytest environment
with a pyfakefs present (which of course we have in our unit tests).
We can work around this by calling add_real_directory() to add the real
source directory to the fakefs, so the os.path.isfile() test will pass.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 import json
 import pytest
 
+import Cryptodome
 import yaml
 from mock import patch
 from pyfakefs.fake_filesystem_unittest import TestCase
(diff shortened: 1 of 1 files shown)


tserong:
Let's see if this passes...

tserong:
OK, uhh... Kinda? At least it fixes the "tox / pytest" checks, but not the "Testing / build" checks (compare with https://github.com/ceph/ceph-salt/pull/461 which is failing all of the above)

mgfritch:
@susebot run teuthology

mgfritch:
@susebot run contiguous-integration

mgfritch:
jenkins retest this please

smithfarm (approved):

smithfarm:
Thanks, @tserong and @mgfritch !

PR #465, opened 2021-11-08
ceph-salt-formula: give explicit IP addr when adding host

### LINKED ISSUE #463: ceph-salt can no longer bootstrap pacific as of ceph commit 5c9227bbdf987365ae90433ac4e501e1a88b7aca
As of https://github.com/ceph/ceph/commit/5c9227bbdf98 ceph-salt can no longer bootstrap pacific clusters:

```
    master: [2021-11-08 09:44:23.138464] [master.pacific-m] Finished with failures
    master: Failure in minion: master.pacific-mini.test
    master: __id__: add host to ceph orch
    master: __run_num__: 104
    master: __sls__: ceph-salt.apply.cephorch
    master: changes: {}
    master: comment: "Warning: Permanently added 'master.pacific-mini.test' (ECDSA) to the list\
    master:   \ of known hosts.\r\nError EINVAL: Can not automatically resolve ip address of host\
    master:   \ where active mgr is running. Please explicitly provide the address."
    master: duration: 748.424
    master: name: add host to ceph orch
    master: result: false
    master: start_time: '09:44:22.328683'
    master: state: ceph_orch_|-add host to ceph orch_|-add host to ceph orch_|-add_host
    master: Finished execution of ceph-salt formula
```

ANALYSIS:

The relevant yaml is:

```
{{ macros.begin_stage('Add host to ceph orchestrator') }}
add host to ceph orch:
  ceph_orch.add_host:
    - host: {{ my_hostname }}
    - is_admin: {{ is_admin }}
    - failhard: True
{{ macros.end_stage('Add host to ceph orchestrator') }}
```

which triggers the following Salt State code:

and `my_hostname` is getting populated by this line:

```
{% set my_hostname = salt['ceph_salt.hostname']() %}
```

which in turns triggers the following function:

```
def hostname():
    return socket.gethostname()
```

All this seems to mean that the hostname returned by `socket.gethostname()` is passed to the following function, `add_host`:

```
def add_host(name, host, is_admin=False):
    """ 
    Requires the following grains to be set:
      - ceph-salt:execution:admin_host
    """
    ret = {'name': name, 'changes': {}, 'comment': '', 'result': False}
    admin_host = __salt__['grains.get']('ceph-salt:execution:admin_host')
    cmd_ret = __salt__['ceph_salt.ssh'](
                       admin_host,
                       "sudo ceph orch host add {}{}".format(
                           host,
                           ' --labels _admin' if is_admin else '', 
                        ),  
                       attempts=10)
    if cmd_ret['retcode'] == 0:
        ret['result'] = True
    else:
        ret['comment'] = cmd_ret.get('stderr')
    return ret 
```

Now, when I try `socket.gethostname()` in the sesdev environment, it returns the short hostname (`master`). And it seems that cephadm is expecting to get the short hostname here. But I couldn't find any log message to confirm that ceph-salt is really sending the short hostname.

smithfarm:
The upstream cephadm developers recently added a commit - https://github.com/ceph/ceph/commit/0facfac91fd8f71e5a8b869d818e7c2b07b93516 - which was backported to pacific and was found to be causing our post-bootstrap "ceph orch host add" command to fail because we weren't specifying the IP address explicitly.

While the developers are now discussing whether the command really should fail in this way, it seems prudent to give our IP address here explicitly, since mgr/cephadm is going to try to resolve the hostname, anyway.

Fixes: https://github.com/ceph/ceph-salt/issues/463

### DIFF
--- a/ceph-salt-formula/salt/_modules/ceph_salt.py
+++ b/ceph-salt-formula/salt/_modules/ceph_salt.py
@@ -184,6 +184,10 @@ def hostname():
     return socket.gethostname()
 
 
+def ip_address():
+    return socket.gethostbyname(socket.gethostname())
+
+
 def probe_fqdn():
     """
     Returns 'YES' if FQDN environment detected, 'NO' if not detected, or 'FAIL'
--- a/ceph-salt-formula/salt/_states/ceph_orch.py
+++ b/ceph-salt-formula/salt/_states/ceph_orch.py
@@ -80,7 +80,7 @@ def wait_until_ceph_orch_available(name, timeout=1800):
     ret['result'] = True
     return ret
 
-def add_host(name, host, is_admin=False):
+def add_host(name, host, ipaddr, is_admin=False):
     """
     Requires the following grains to be set:
       - ceph-salt:execution:admin_host
@@ -89,9 +89,10 @@ def add_host(name, host, is_admin=False):
     admin_host = __salt__['grains.get']('ceph-salt:execution:admin_host')
     cmd_ret = __salt__['ceph_salt.ssh'](
                        admin_host,
-                       "sudo ceph orch host add {}{}".format(
+                       "sudo ceph orch host add {} {}{}".format(
                            host,
-                           ' --labels _admin' if is_admin else '',
+                           ipaddr,
+                           ' _admin' if is_admin else '',
                         ),
                        attempts=10)
     if cmd_ret['retcode'] == 0:
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephorch.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephorch.sls
@@ -3,12 +3,14 @@
 {% if 'cephadm' in grains['ceph-salt']['roles'] %}
 
 {% set my_hostname = salt['ceph_salt.hostname']() %}
+{% set my_ipaddr = salt['ceph_salt.ip_address']() %}
 {% set is_admin = 'admin' in grains['ceph-salt']['roles'] %}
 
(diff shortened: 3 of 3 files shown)


smithfarm:
@trociny

Daniel-Pivonka (approved):

mgfritch (approved):

tserong (commented):
There's an unavoidable kink with `socket.gethostbyname(socket.gethostname())`: if you happen to have specified the system hostname with the loopback address in /etc/hosts, this construct will return 127.0.0.1. I'm almost certain this is unavoidable, and you shouldn't be configuring /etc/hosts like that anyway, but, it might be worth checking if the address returned starts with "127.", and aborting or erroring out in that case. Otherwise looks fine.

tserong:
...I realise bailing out in this context may be problematic, so don't let my comment stop this PR. Might be worth opening an issue to somehow add a hostname resolution check elsewhere in ceph-salt, as a general sanity check?

sebastian-philipp:
alternative would be to have the check in https://github.com/ceph/ceph/pull/43856

smithfarm:
> alternative would be to have the check in [ceph/ceph#43856](https://github.com/ceph/ceph/pull/43856)

Well, it's not an "either-or" proposition. It's possible (and probably even desirable) for both ceph-salt and mgr/cephadm to implement this same check.

But ceph-salt has an advantage: it isn't running in a container. So it doesn't get confused about its own IP address like mgr/cephadm does.

smithfarm:
Anyway, let's move the discussion of the IP address sanity check to #466

I'm going to merge this soon if there are no further comments.

PR #467, opened 2021-11-16
Move ceph-salt-registry-json creation outside bootstrap

tserong:
Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1187015

### DIFF
--- a/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
+++ b/ceph-salt-formula/salt/ceph-salt/apply/cephbootstrap.sls
@@ -18,18 +18,6 @@
 
 {{ macros.begin_step('Login into registry') }}
 
-create ceph-salt-registry-json:
-  file.managed:
-    - name: /tmp/ceph-salt-registry-json
-    - source:
-        - salt://ceph-salt/files/registry-login-json.j2
-    - template: jinja
-    - user: root
-    - group: root
-    - mode: '0600'
-    - backup: minion
-    - failhard: True
-
 login into registry:
   cmd.run:
     - name: |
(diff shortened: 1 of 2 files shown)


sebastian-philipp:
Do you want to replace things with `--registry-json`? https://docs.ceph.com/en/latest/cephadm/install/?highlight=registry%20login#further-information-about-cephadm-bootstrap

smithfarm:
For me, it's confusing to have something in "cephbootstrap.sls" that has been "moved outside bootstrap". Syntactically speaking, I understand that it will probably work that way, but just in the interests of having meaningful names, I would suggest to move the code into a completely different sls file. It could be called "cephsaltregistryjson.sls", for example.

tserong:
> Do you want to replace things with `--registry-json`

Ah, yeah, looks like we've missed one of them. We do currently have `cephadm bootstrap --registry-json /tmp/ceph-salt-registry-json` and `ceph cephadm registry-login -i /tmp/ceph-salt-registry-json`, but the earlier `cephadm registry-login` is still using the old form where it passes registry, username and password via the CLI. I'll fix that too.

> For me, it's confusing to have something in "cephbootstrap.sls" that has been "moved outside bootstrap". Syntactically speaking, I understand that it will probably work that way, but just in the interests of having meaningful names, I would suggest to move the code into a completely different sls file. It could be called "cephsaltregistryjson.sls", for example.

Totally fair, will rejig.

tserong:
Alright, that ought to be a bit cleaner.

sebastian-philipp:
https://github.com/ceph/ceph/pull/43889#issuecomment-970231254

tserong:
(Jenkins failure looks to me like Jenkins itself is broken - see also https://ceph-ci.suse.de/blue/organizations/jenkins/sesdev-integration/detail/master/1095/pipeline/, assuming I've got that URL correct)

smithfarm (approved):

PR #471, opened 2022-02-23
ceph_salt/core: treat 127.0.1.1 as loopback too

kshtsk:
Handle the case when interface has bound to IP address 127.0.1.1
instead of 127.0.0.1, which can also happen.

Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.com>


mgfritch (commented):
see #472, let's use the `ipaddress` module for this :)

kshtsk:
> see #472, let's use the `ipaddress` module for this :)

thanks, will close this as a duplication

PR #472, opened 2022-02-23
use ipaddress module to determine loopback interfaces

mgfritch:
handling for other loopback address variants (127.0.1.1, ::1, etc.)

### DIFF
--- a/ceph_salt/core.py
+++ b/ceph_salt/core.py
@@ -1,8 +1,9 @@
+import base64
 import copy
+import hashlib
+import ipaddress
 import logging
 import os
-import base64
-import hashlib
 
 from Cryptodome.PublicKey import RSA
 import salt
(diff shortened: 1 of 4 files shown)


ceph_salt/core.py
```diff
@@ -44,18 +45,21 @@ def ipsv6(self):
 
     @property
     def public_ip(self):
+        def _is_loopback(addr):
+            return ipaddress.ip_address(addr).is_loopback
```
kshtsk, line 49: Cool, this looks covers 127.0.1.1 too:
```
>>> ipaddress.ip_address('127.0.1.1').is_loopback
True
```

ceph_salt/core.py
```diff
@@ -44,18 +45,21 @@ def ipsv6(self):
...
+            if _is_loopback(_public_ip):
+                logger.debug('fqdn_ipv4 grain is a loopback address, falling back to ipv4 grain')
                 result = GrainsManager.get_grain(self.minion_id, 'ipv4')
                 for addr in result[self.minion_id]:
-                    if addr != '127.0.0.1':
+                    if not _is_loopback(addr):
                         _public_ip = addr
                         break
-                if _public_ip == '127.0.0.1':
-                    logger.warning("'%s' public IP is the loopback interface IP ('127.0.0.1')",
```
kshtsk, line 58: Would be great if you add _public_ip value to the log instead of 127.0.0.1

ceph_salt/core.py
```diff
@@ -44,18 +45,21 @@ def ipsv6(self):
...
     @property
     def public_ip(self):
+        def _is_loopback(addr):
+            return ipaddress.ip_address(addr).is_loopback
+
         if self._public_ip is None:
             result = GrainsManager.get_grain(self.minion_id, 'fqdn_ip4')
             _public_ip = result[self.minion_id][0]
-            if _public_ip == '127.0.0.1':
-                logger.debug('fqdn_ipv4 grain is 127.0.0.1, falling back to ipv4 grain')
```
kshtsk, line 51: For better debugging I guess it is better to print the _public_ip here the same way as it was before

kshtsk (approved):
Looks good, but minor debugging/logging improvements required

ceph_salt/core.py
mgfritch: yeah good point, I've re-added the `_public_ip` to the log message

ceph_salt/core.py
mgfritch: added `_public_ip` as part of the log message

tserong (approved):
Nicely done

PR #482, opened 2022-04-13
Fix `ceph-salt update` when run prior to cluster deployment

tserong:
Three changes here:

1) Ensure cephadm package is installed during `ceph-salt update`
2) Only add _admin host labels during update if cluster is deployed
3) Add diagnostic hint in case of ceph-salt-formula errors

This'll want backporting to the octopus branch too.


mgfritch (approved):
lgtm!

PR #484, opened 2022-05-17
Usage of daemon_name in stop_service

giubacc:
stop_service function was incorrectly using the service_type
when stopping a service.
Now the actual daemon_name(s) is/are extracted and the value(s)
used to perform a service's stop.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1199488
Signed-off-by: Giuseppe Baccini <giuseppe.baccini@suse.com>


mgfritch (approved):
lgtm!

PR #487, opened 2022-12-19
Bump to v16.2.4, fix a couple of things

tserong:
Two fixes here:

* Run `saltutil.sync_runners` in ceph-salt-formula %post script
* Don't explicitly install ceph-common (rely on ceph-base dependency)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         py:
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   pytest:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         include:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [16.2.4] - 2022-12-19
 ### Fixed
+- Use daemon name rather than service type when stopping services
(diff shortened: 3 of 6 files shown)


tserong:
@m-ildefons are you interested in reviewing this? I've got a build on IBS if you want to play (`sesdev create ses7p --product --repo-priority --repo http://dist.suse.de/ibs/home:/tserong:/branches:/SUSE:/SLE-15-SP3:/Update:/Products:/SES7:/Update/standard` should do the right thing)

tserong:
(Note to reviewers: the "Use daemon name rather than service type when stopping services" line in the changelog is a reference to https://github.com/ceph/ceph-salt/pull/484 which was merged earlier in the year but not yet included in a release)

tserong:
Jenkins failure appears to be due to broken infra, not this PR

mgfritch (approved):

tserong:
Sorry @mgfritch, do you mind having another look? I've just pulled in another fix...

tserong:
(extra force pushes to deal with lint issues - sorry for the noise)

.github/workflows/testing.yml
```diff
@@ -9,7 +9,6 @@ jobs:
     strategy:
       matrix:
         py:
-          - 3.6
```
m-ildefons: Rather than remove 3.6 outright, why not change the GH worker instance to be an `ubuntu-20.04`? Given that when run in anger, this most likely will be running on Python 3.6 (default on SLE 15 SP3)?

.github/workflows/tox.yml
```diff
@@ -9,8 +9,6 @@ jobs:
     strategy:
       matrix:
         include:
-        - tox-env: py36
-          py-ver: 3.6
```
m-ildefons: See above.

m-ildefons (changes requested):
Other than the Python 3.6 thing, this looks good to me

m-ildefons:
The packages on IBS work fine for me, so only the Python 3.6 question remains.

.github/workflows/testing.yml
tserong: Excellent idea, thanks!

mgfritch (approved):

